PR #24, opened 2019-11-20
Issue #SB-15699 feat: Updating objectType from small case to camel case

amitpriyadarshi:


maheshkumargangula:
@amitpriyadarshi - I'm merging this but, I need another pull-request with unit test cases to validate the case sensitivity.

PR #34, opened 2019-11-27
Partial update hierarchy

PradyumnaNagendra:


learning-api/hierarchy-service/src/main/scala/org/sunbird/utils/NodeUtil.scala
```diff
@@ -0,0 +1,81 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import org.apache.commons.collections4.CollectionUtils
+import org.sunbird.graph.dac.model.{Node, Relation}
+import org.sunbird.graph.schema.DefinitionNode
+
+import scala.collection.JavaConverters._
+
+object NodeUtil {
+    val mapper: ObjectMapper = new ObjectMapper()
```
maheshkumargangula: Create `ScalaJsonUtil` and use it across.

PR #66, opened 2019-12-06
Issue #SC-0000 fix: Fixed Invalid Version Key Issue

krgauraw:


ontology-engine/graph-engine_2.11/src/main/scala/org/sunbird/graph/schema/DefinitionNode.scala
```diff
@@ -72,6 +73,10 @@ object DefinitionNode {
             resetJsonProperties(dbNode, graphId, version, schemaName)
             val inputNode: Node = definition.getNode(dbNode.getIdentifier, request.getRequest, dbNode.getNodeType)
             setRelationship(dbNode,inputNode)
+            if (dbNode.getIdentifier.endsWith(".img") && StringUtils.equalsAnyIgnoreCase("Yes", dbNode.getMetadata.get("isImageNodeCreated").asInstanceOf[String])) {
+                inputNode.getMetadata.put("versionKey", dbNode.getMetadata.get("versionKey"))
+                dbNode.getMetadata.remove("isImageNodeCreated")
+            }
```
maheshkumargangula: Why don't you put this logic in VersioningNode.scala?

PR #73, opened 2019-12-12
gethierarchy implemented

neha0305verma:


learning-api/hierarchy-manager/src/main/scala/org/sunbird/managers/HierarchyManager.scala
```diff
@@ -123,7 +173,8 @@ object HierarchyManager {
 
     def fetchHierarchy(request: Request)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
         val req = new Request(request)
-        req.put("identifier", request.get("rootId").asInstanceOf[String] + imgSuffix)
+        if (!StringUtils.isEmpty(request.get("mode").asInstanceOf[String]) && request.get("mode").equals("edit")) req.put("identifier", request.get("rootId").asInstanceOf[String] + imgSuffix)
```
PradyumnaNagendra: Since `mode` is introduced here, update addHierarchy and removeHierarchy hierarchy to include mode by default

PR #78, opened 2019-12-16
Issue #SC-1557 feat: Added New Cache Module

krgauraw:


platform-core/platform-cache/src/main/scala/org/sunbird/cache/impl/RedisCache.scala
```diff
@@ -0,0 +1,182 @@
...
+	 * @param key
+	 * @param handler
+	 * @return String
+	 */
+	def get(key: String, handler: (String, String) => String = defaultHandler): String = {
+		val jedis = getConnection
+		try {
+			var data = jedis.get(key)
+			if (null!=handler && (null == data || data.isEmpty))
+				data = handler(key, key)
```
maheshkumargangula: If the data from the handler is not empty, we should set it to Redis.

krgauraw: done. Please Review

PR #89, opened 2019-12-27
Issue #SB-16894 feat: added itemset apis implementation

krgauraw:


schemas/itemset/2.0/schema.json
```diff
@@ -22,7 +22,8 @@
             "type": "number"
         },
         "type": {
-            "type": "string"
+            "type": "string",
```
maheshkumargangula: Add enum with `materialised` otherwise, the system accepts any string.

krgauraw: done

PR #169, opened 2020-03-09
SB-18218 - Channel APIs Refactoring

rheafernandes:


assessment-api/assessment-service/conf/application.conf
```diff
@@ -310,3 +310,5 @@ languageCode {
   telugu : "te"
 }
 
+language.graph_ids=["as","bn","en","gu","hi","hoc","jun","ka","mai","mr","unx","or","san","sat","ta","te","urd"]
```
maheshkumargangula: We don't need this config. Remove it.

content-api/content-actors/src/main/scala/org/sunbird/channel/managers/ChannelManager.scala
```diff
@@ -0,0 +1,88 @@
...
+      else {
+        val licenseList = searchLicenseList()
+        if (licenseList.contains(request.getRequest.get(ChannelConstants.DEFAULT_LICENSE))) {
+          RedisCache.saveList(ChannelConstants.LICENSE_REDIS_KEY, licenseList)
+          response
+        } else
+          throw new ResourceNotFoundException("LICENSE_NOT_FOUND", "License " + request.getRequest.get(ChannelConstants.DEFAULT_LICENSE).asInstanceOf[String] + " does not exist")
+      }
+    } else
+      response
```
maheshkumargangula: We should use "property as relation" behaviour to implement this. No need to explicitly validate it.

PR #234, opened 2020-04-02
Issue #SC-1700 feat: search-service refactoring

PradyumnaNagendra:


search-api/search-core/pom.xml
```diff
@@ -90,6 +90,11 @@
             <artifactId>elasticsearch-rest-high-level-client</artifactId>
             <version>6.2.4</version>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+        </dependency>
```
maheshkumargangula, line 97: add test scope for this dependency.

PradyumnaNagendra: done

PR #245, opened 2020-04-09
Changes made to positioning of filepath in url generation

rheafernandes:


amitpriyadarshi:
@rheafernandes circleci is failing because of wrong h5p file url. I think you have not update yor fork with 2.10.0 origin.

PR #274, opened 2020-05-13
SH-9: added courseType for backward compatibility

krgauraw:


search-api/search-actors/src/main/java/org/sunbird/actors/SearchActor.java
```diff
@@ -104,6 +104,16 @@ private SearchDTO getSearchDTO(Request request) throws Exception {
             if (filters.containsKey("relatedBoards"))
                 filters.remove("relatedBoards");
 
+            if (filters.containsKey(SearchConstants.contentType) && null != filters.get(SearchConstants.contentType)) {
```
amitpriyadarshi, line 107: Gaurav, in case if we dont pass filter, first line will fail.
filters.containsKey(SearchConstants.contentType)

search-api/search-actors/src/main/java/org/sunbird/actors/SearchActor.java
```diff
@@ -104,6 +104,16 @@ private SearchDTO getSearchDTO(Request request) throws Exception {
...
+                List<String> contentTypeList = getList(filters.get(SearchConstants.contentType));
+                Object courseTypeFromFilter = filters.get(SearchConstants.courseType);
```
amitpriyadarshi, line 109: If contentTypeList is not empty and contentTypeList has only one value i.e. `Course`, then only we will capture `courseType`

search-api/search-actors/src/main/java/org/sunbird/actors/SearchActor.java
krgauraw: that won't happen. please see full code.

search-api/search-actors/src/main/java/org/sunbird/actors/SearchActor.java
krgauraw: I don't see any impact on reading courseType before that condition. if we put this under such condition, it will result into nested control instruction.

krgauraw: code refactored as per above comment.

search-api/search-actors/src/main/java/org/sunbird/actors/SearchActor.java
```diff
@@ -104,6 +104,18 @@ private SearchDTO getSearchDTO(Request request) throws Exception {
             if (filters.containsKey("relatedBoards"))
                 filters.remove("relatedBoards");
 
+            if (filters.containsKey(SearchConstants.contentType) && null != filters.get(SearchConstants.contentType)) {
+                List<String> contentTypeList = getList(filters.get(SearchConstants.contentType));
+                if ((CollectionUtils.isNotEmpty(contentTypeList) && contentTypeList.size() == 1) && contentTypeList.contains("Course")) {
```
maheshkumargangula, line 109: The value check should not be case sensitive.

1. Course
2. course
3. COURSE
All are valid for search API.

krgauraw: changes done. 
https://github.com/project-sunbird/knowledge-platform/pull/276/files

PR #285, opened 2020-05-28
Added fix to null hierarchy, which was deleting existing hierarchy

rheafernandes:


maheshkumargangula:
@rheafernandes - Merging the pull-request. But, expecting another pull-request with test cases ASAP.

PR #330, opened 2020-07-16
Issue #DP-875: feat- Added implementation to handle composed h5p file

rheafernandes:


platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/H5PMimeTypeMgrImpl.scala
```diff
@@ -31,10 +37,13 @@ class H5PMimeTypeMgrImpl(implicit ss: StorageService) extends BaseMimeTypeManage
...
 
-    override def upload(objectId: String, node: Node, fileUrl: String, filePath: Option[String])(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
+    override def upload(objectId: String, node: Node, fileUrl: String, filePath: Option[String], params: UploadParams)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
         validateUploadRequest(objectId, node, fileUrl)
         val file = copyURLToFile(objectId, fileUrl)
-        upload(objectId, node, file, filePath)
+        if (StringUtils.equalsIgnoreCase(params.fileFormat, COMPOSED_H5P_ZIP))
+            Future(Map[String, AnyRef]("identifier" -> objectId, "artifactUrl" -> fileUrl, "downloadUrl" -> fileUrl, "size" -> getFileSize(file).asInstanceOf[AnyRef]))
+        else
+            upload(objectId, node, file, filePath, params)
```
maheshkumargangula, line 46:
```suggestion
            upload(objectId, node, file, filePath, params)
```
Use the upload method only. We should use the local storage for the content.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/models/UploadParams.scala
```diff
@@ -0,0 +1,26 @@
+package org.sunbird.models
+
+import play.api.mvc.QueryStringBindable
+
+
+case class UploadParams(fileFormat: String = "", validation: Boolean = true)
+
+object UploadParams {
```
maheshkumargangula, line 8: Move this object to the content-service module.

platform-modules/mimetype-manager/pom.xml
```diff
@@ -80,6 +84,17 @@
...
+            <groupId>com.typesafe.play</groupId>
+            <artifactId>play_${scala.major.version}</artifactId>
+            <version>${play2.version}</version>
+            <exclusions>
+                <exclusion>
+                    <artifactId>guava</artifactId>
+                    <groupId>com.google.guava</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
maheshkumargangula, line 97: remove this dependency.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/H5PMimeTypeMgrImpl.scala
rheafernandes: done

platform-modules/mimetype-manager/pom.xml
rheafernandes: removed query string binder

platform-modules/mimetype-manager/src/main/scala/org/sunbird/models/UploadParams.scala
rheafernandes: removed query string binder

PR #428, opened 2020-09-28
SC-1955 Added testcase for copy content and upload content

rheafernandes:


platform-modules/mimetype-manager/src/test/scala/org/sunbird/cloudstore/StorageServiceTest.scala
```diff
@@ -0,0 +1,5 @@
+package org.sunbird.cloudstore
```
PradyumnaNagendra, line 1: remove this if not required

rheafernandes: fixed it

PradyumnaNagendra (approved):

PR #541, opened 2021-02-19
SC-2226 Event and Event Set APIs

univermal:


univermal:
Will be merged to 3.8.0

PR #545, opened 2021-02-24
Add `printable` attribute to collection schema

bharatkashyap:


maheshkumargangula (approved):

maheshkumargangula:
@bharatkashyap - you can update the schema of the "object primary category definition" to handle this use case. No need to introduce it in at core object level. So, closing this PR.
cc: @pallakartheekreddy @vrayulu

PR #789, opened 2022-02-17
Added code to check config param for Index file name

karthik-tarento:
We require an optional way to check the HTML resource entry file instead of expecting "index.html" as file name.
Hence, introduced the code to check configuration and decide mandating "index.html" file is required or not.

The configuration can be added in knowledge-platform/content-api/content-service/conf/application.conf to enable / disable and missing out the configuration would be treated as "index.html" file is mandatory.


karthik-tarento:
New PR - #817 raised for latest version

PR #819, opened 2022-04-08
SB-29495 | schema changes for iGOT

karthik-tarento:


karthik-tarento:
@AmiableAnil -- Can you review this PR and provide your comments ?

AmiableAnil:
@karthik-tarento Why the iGOT specific schema changes are required? It can we iGOT can have their own schema in the forked repo.

karthik-tarento:
> @karthik-tarento Why the iGOT specific schema changes are required? It can we iGOT can have their own schema in the forked repo.

@AmiableAnil -- we are moving away from forked repository since iGot will be contributing and using SB releases directly from future releases. Hence we would like to make sure the schema changes in SB repo too.

pramodkvarma:
I hardly see any PR comments, why is this blocked @karthik-tarento
If core services (API code) is enhanced using these additional parameters, both enhancement and this have to be merged in. @vrayulu @vinukumar-vs

vinukumar-vs (approved):

vrayulu:
@vinukumar-vs @karthik-tarento - are the schema changes discussed before making them part of the default definitions in the knowlg repo? There are attributes like "taxonomyPaths_v2", "competencies_v3" which are having version numbers, what does these version numbers mean? 
and the attributes "choices" and "rhsChoices" are added to question schema, but the QuML spec defines the options/choices using the "interactions" attribute. Would this not be a deviation from QuML spec?

And as Pramod asked, is there any behaviour/code change related to these attributes? If yes, we need that code also to be contributed back.

@vinukumar-vs - we also discussed about each adopter managing their own schema definitions (in their repositories) and deploy it along it with the source code from the core sunbird repository. Is that approach considered in this case and if yes, what is the reason for not taking that approach?

vinukumar-vs:
@vrayulu As I mentioned over the mail, these schema changes doesn't have any core code changes on SB-Knowledge. As I discussed with @pkranga, these attributes are used on Sunbird-CB to drive some behavior(Which we were not aware).  

Hence I was suggesting it is specific to the adopter & implementation/behavior is also on the adopter side, then we will not take those schema changes as part of SB-Knowlg. But as discussed with @pkranga, Subird-CB related schema changes are we should consider, because it is also Open-Source & it should leverage the SB-Knowlg core schema similar to Sunbird-ED.

@pkranga @karthik-tarento You have some specific reason to append the versions for attributes. Can you please explain? I am not able to recollect it fully.

pramodkvarma:
Attributes that are not used by code (core services) should be layered on top. All the parameters that are used by core code can be retained within core config. 

For now, go ahead. But, add an issue tracker item @pkranga and make it a separate config that can be merged during load/init by CB layer.

PR #945, opened 2023-06-05
Issue #KN-848 feat: Scala version update release-5.5.0

ShouraySolanki:


platform-core/schema-validator/src/main/java/org/sunbird/schema/impl/JsonSchemaValidator.java
```diff
@@ -37,7 +37,7 @@ private void loadSchema() throws Exception {
             InputStream stream = new URL( basePath + "schema.json").openStream();
             this.schema = readSchema(stream);
         }else {
-            Path schemaPath = new File( basePath + "schema.json").toPath();
+            Path schemaPath = new File( basePath + "config.json").toPath();
```
maheshkumargangula, line 40: Why did you change this?

maheshkumargangula (changes requested):

ShouraySolanki:
I will revert it back for content i didn't see the schema.json file so
changed it check if the path issue is due to this file. But still it was
failing.

PR #1191, opened 2026-02-26
refactor: replace println/System.out.println with TelemetryManager lo…

pallakartheekreddy:
…gging

Replace all raw println and System.out.println calls in business logic with structured TelemetryManager calls (debug/error/warn) so that log output is captured by the platform's logging pipeline and not silently dropped in production.

Files changed:
- ContentActor.scala: warn for missing default licence
- ChannelManager.scala: error for category fetch failure (added import)
- NodeValidator.scala: debug for singular identifier lookup (added import)
- DataSubGraph.scala: debug for subgraph traversal diagnostics (added import)
- StorageService.scala: error for getUri exception (added import)
- CollectionTOCUtil.scala: error for validateDIALCodes exception
- ObjectCategoryDefinitionActor.scala: debug for fallback _all lookup (added import)

### DIFF
--- a/ontology-engine/graph-engine_2.13/src/main/scala/org/sunbird/graph/path/DataSubGraph.scala
+++ b/ontology-engine/graph-engine_2.13/src/main/scala/org/sunbird/graph/path/DataSubGraph.scala
@@ -9,6 +9,7 @@ import org.sunbird.graph.dac.model.{Node, Relation, SubGraph}
 import org.sunbird.graph.nodes.DataNode
 import org.sunbird.graph.schema.{DefinitionFactory, DefinitionNode, ObjectCategoryDefinition}
 import org.sunbird.graph.utils.NodeUtil
+import org.sunbird.telemetry.logger.TelemetryManager
 import scala.jdk.CollectionConverters._
 import org.sunbird.graph.utils.NodeUtil.{convertJsonProperties, handleKeyNames}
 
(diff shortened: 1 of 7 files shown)


ontology-engine/graph-engine_2.13/src/main/scala/org/sunbird/graph/path/DataSubGraph.scala
```diff
@@ -107,7 +108,7 @@ object DataSubGraph {
         finalMetadata.keySet.retainAll(fields)
       finalMetadata.put("identifier", node.getIdentifier)
     }
-    println("definitionMap  "+ definitionMap)
+    TelemetryManager.debug("definitionMap  " + definitionMap)
```
maheshkumargangula:
```suggestion
```

ontology-engine/graph-engine_2.13/src/main/scala/org/sunbird/graph/path/DataSubGraph.scala
```diff
@@ -37,7 +38,7 @@ object DataSubGraph {
     val dataMap = new util.HashMap[String, AnyRef]
     val relMap = new util.HashMap[String, AnyRef]
     readSubGraphData(request, dataMap, relMap).map(sub => {
-      println("subGraphData out " + sub)
+      TelemetryManager.debug("subGraphData out " + sub)
```
maheshkumargangula:
```suggestion
```

maheshkumargangula (changes requested):

maheshkumargangula (approved):

PR #1234, opened 2026-05-11
feat: Implement SCORM package ingestion, playback, and integrity validation

romitshah02:
Summary of Change
Implemented end-to-end SCORM package ingestion and playback support. This includes backend ingestion logic (ScormMimeTypeMgrImpl), editor support for MIME detection and binary upload, and a SCORM API shim in the player. Included an integrity check to verify the existence of the launchFile specified in imsmanifest.xml.

Motivation
To enable compatibility with standard SCORM learning modules within the platform, ensuring robust package ingestion and reliable playback.

Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

How Has This Been Tested?
- [x] Verified E2E flow (Authoring -> Ingestion -> Playback) locally.
- [x] Verified invalid SCORM packages now throw ClientException during ingestion.
- [x] Verified valid SCORM packages ingest and playback correctly.

Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * SCORM support: package upload, extraction, artifact storage, and automatic launch-file detection; review and publish flows.

* **Bug Fixes / Validation**
  * Rejects invalid SCORM packages and missing/invalid launch files with clear errors.

* **Security**
  * Extraction hardened to prevent path traversal/Zip-Slip and ensure safe launch-file resolution.

* **Tests**
  * Added unit tests covering SCORM upload validation and failure cases.

* **Schema**
  * Content schema now allows the SCORM MIME type.

* **Chores / Dev**
  * Local seed-data tooling and README updates.

* **Infrastructure**
  * Compose/VM and CI updates: OpenSearch migration and simplified build workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sunbird-Knowlg/knowledge-platform/pull/1234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+
+class ScormMimeTypeMgrImpl(implicit ss: StorageService) extends BaseMimeTypeManager()(ss) with MimeTypeManager {
+
+    override def upload(objectId: String, node: Node, uploadFile: File, filePath: Option[String], params: UploadParams)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
+        validateUploadRequest(objectId, node, uploadFile)
+        TelemetryManager.info("SCORM content upload for objectId:: " + objectId)
+
+        if (isValidPackageStructure(uploadFile, List("imsmanifest.xml"))) {
+            val extractionBasePath = getBasePath(objectId)
+            extractPackage(uploadFile, extractionBasePath)
```
sntiwari1: **Resource Leak: Temp directory is never cleaned up**

The `extractionBasePath` directory is created via `getBasePath(objectId)` and the package is extracted into it, but the directory is never deleted — not on the happy path, not on exception. This leaks disk space on every upload.

Compare with `H5PMimeTypeMgrImpl` line 41 which explicitly deletes the zip file after use. The extracted directory here is more expensive to leave around.

Suggested fix:
```scala
val extractionBasePath = getBasePath(objectId)
try {
  extractPackage(uploadFile, extractionBasePath)
  // ... validation and upload ...
  Future { ... }
} finally {
  delete(new File(extractionBasePath))
}
```
The `delete` helper method is already available in `BaseMimeTypeManager`.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+            val manifestFile = new File(extractionBasePath + File.separator + "imsmanifest.xml")
+            val xml = scala.xml.XML.loadFile(manifestFile)
```
sntiwari1: **Security: scala.xml.XML.loadFile is vulnerable to XXE (XML External Entity) injection**

`scala.xml.XML.loadFile` uses the default Java SAX parser with external entity resolution enabled. A maliciously crafted `imsmanifest.xml` inside the uploaded zip could include an XXE payload to read local files from the server or trigger server-side request forgery (SSRF).

This is a medium-severity server-side vulnerability because:
1. The manifest is extracted from a user-uploaded zip before being parsed
2. The default SAX parser resolves `DOCTYPE` and external entity declarations

Mitigation — disable external entity resolution explicitly:
```scala
import javax.xml.parsers.SAXParserFactory
val spf = SAXParserFactory.newInstance()
spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true)
spf.setFeature("http://xml.org/sax/features/external-general-entities", false)
spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false)
val parser = spf.newSAXParser()
val xml = scala.xml.XML.withSAXParser(parser).loadFile(manifestFile)
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+class ScormMimeTypeMgrImpl(implicit ss: StorageService) extends BaseMimeTypeManager()(ss) with MimeTypeManager {
+
+    override def upload(objectId: String, node: Node, uploadFile: File, filePath: Option[String], params: UploadParams)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
+        validateUploadRequest(objectId, node, uploadFile)
+        TelemetryManager.info("SCORM content upload for objectId:: " + objectId)
+
+        if (isValidPackageStructure(uploadFile, List("imsmanifest.xml"))) {
+            val extractionBasePath = getBasePath(objectId)
+            extractPackage(uploadFile, extractionBasePath)
+            val manifestFile = new File(extractionBasePath + File.separator + "imsmanifest.xml")
```
sntiwari1: **Security: Zip Slip vulnerability via extractPackage**

The `extractPackage` call on the line above (and the shared `BaseMimeTypeManager.extractPackage` implementation) does **not** validate that extracted entry paths remain within the target directory. A malicious archive could contain entries like:

```
../../etc/passwd
../../../app/conf/application.conf
```

These traverse outside `extractionBasePath` and overwrite arbitrary server files.

The vulnerable code in `BaseMimeTypeManager`:
```scala
val path = Paths.get(basePath + File.separator + entry.getName)  // no normalization!
```

Fix: normalize and validate every entry path before extraction:
```scala
val targetDir = new File(basePath).getCanonicalFile
val path = Paths.get(basePath + File.separator + entry.getName).normalize()
if (!path.startsWith(targetDir.toPath))
  throw new ClientException("ERR_INVALID_FILE", "Zip entry escapes target directory: " + entry.getName)
```

Note: this is a defect in the shared `BaseMimeTypeManager` that affects all archive-based managers (HTML, H5P, ECML, Plugin), but SCORM is the newest addition and a good opportunity to fix or at minimum document the known risk.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+            val xml = scala.xml.XML.loadFile(manifestFile)
+            // Use local name to ignore namespaces
+            val launchFile = (xml \\ "_").find(node => node.label == "resource" && (node \@ "href").nonEmpty)
+              .map(_ \@ "href")
+              .getOrElse("index.html")
+
+            if (!new File(extractionBasePath + File.separator + launchFile).exists()) {
```
sntiwari1: **Security & Logic: launchFile href value is not sanitized**

The `launchFile` string is taken verbatim from the XML `href` attribute and concatenated directly with `extractionBasePath` to form a filesystem path. There are two issues:

1. **Path traversal**: A value of `"../../../etc/passwd"` would cause `new File(extractionBasePath + File.separator + launchFile).exists()` to check (and potentially leak existence of) files outside the extraction directory.

2. **Absolute path injection**: An href value of `"/etc/passwd"` would make `new File("/etc/passwd")` exist unconditionally on Linux, bypassing this validation entirely and letting the upload proceed with a misleading `launchFile` stored in the node's metadata.

3. **The fallback of `"index.html"`** is returned when no `resource` element with an `href` is found. For strict SCORM compliance, a missing resource element should be treated as an invalid package, not silently defaulted.

Suggested fix:
```scala
val rawHref = (xml \\ "_").find(n => n.label == "resource" && (n \@ "href").nonEmpty)
  .map(_ \@ "href")
  .getOrElse(throw new ClientException("ERR_INVALID_FILE", "No resource href found in imsmanifest.xml"))

// Normalize and verify it stays within the extraction directory
val resolvedPath = new File(extractionBasePath, rawHref).getCanonicalFile
val basePath     = new File(extractionBasePath).getCanonicalFile
if (!resolvedPath.toPath.startsWith(basePath.toPath))
  throw new ClientException("ERR_INVALID_FILE", "Invalid launch file path in imsmanifest.xml")
val launchFile = rawHref
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,58 @@
...
+                TelemetryManager.error("ERR_INVALID_FILE:: " + "Launch file defined in imsmanifest.xml does not exist: " + launchFile)
+                throw new ClientException("ERR_INVALID_FILE", "The launch file '" + launchFile + "' specified in imsmanifest.xml is missing from the package!")
+            }
+
+            val urls: Array[String] = uploadArtifactToCloud(uploadFile, objectId, filePath)
+            Future { Map[String, AnyRef]("identifier" -> objectId, "artifactUrl" -> urls(IDX_S3_URL), "size" -> getFileSize(uploadFile).asInstanceOf[AnyRef], "s3Key" -> urls(IDX_S3_KEY), "launchFile" -> launchFile) }
```
sntiwari1: **Performance: Blocking I/O executes on the caller's thread, not wrapped in a blocking-safe Future**

`uploadArtifactToCloud` (line 36) is a synchronous/blocking cloud storage call. It runs on the calling thread *before* the `Future { ... }` block on line 37, which means it blocks whatever thread dispatched this actor message — potentially the Pekko dispatcher.

Compare with how `H5PMimeTypeMgrImpl` handles this: it also calls `uploadArtifactToCloud` synchronously, so this is an existing pattern in the codebase. However, since SCORM packages additionally perform XML parsing and a disk existence check, the total blocking time per upload is higher.

The idiomatic fix for this module is to wrap all blocking work in `Future { blocking { ... } }` using `scala.concurrent.blocking` to hint the dispatcher:
```scala
import scala.concurrent.blocking
Future {
  blocking {
    val urls = uploadArtifactToCloud(uploadFile, objectId, filePath)
    Map[String, AnyRef]("identifier" -> objectId, "artifactUrl" -> urls(IDX_S3_URL), ...)
  }
}
```
At minimum, move `uploadArtifactToCloud` inside the `Future { }` block so all I/O is logically grouped.

platform-modules/mimetype-manager/src/test/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImplTest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import java.io.File
+import java.util
+import scala.concurrent.Future
+
+class ScormMimeTypeMgrImplTest extends AsyncFlatSpec with Matchers with AsyncMockFactory {
+
+    implicit val ss = mock[StorageService]
+
+    it should "Throw Client Exception for null file" in {
```
sntiwari1: **Test Coverage: Only 1 test case — critical paths are completely untested**

The test file has a single test covering only the `null/empty file` validation path. The following scenarios are not tested and need to be added before merge:

| Scenario | Why it matters |
|---|---|
| Valid SCORM zip with correct `imsmanifest.xml` and existing `href` | Happy path — verifies upload succeeds and `launchFile` is in the result map |
| Zip missing `imsmanifest.xml` | Verifies `ClientException("ERR_INVALID_FILE", "Invalid SCORM package: ...")` is thrown |
| `imsmanifest.xml` present but referenced launch file missing from zip | Verifies the new integrity check added in this PR actually works |
| File URL overload delegates to file overload | Verifies the `upload(fileUrl))` path calls the file-based `upload` correctly |
| `review` with no `artifactUrl` on node | Verifies `ClientException("VALIDATOR_ERROR", ...)` is thrown |
| `publish` with valid `artifactUrl` on node | Happy path for publish |

Also: the `implicit val ss = mock[StorageService]` is declared at class scope with `AsyncMockFactory` which resets mock state between tests. In the ApkMimeTypeMgrImplTest pattern, `ss` is declared per-test. Move it inside the test body or use a `beforeEach` reset, or stale expectations from one test will bleed into another.

There are no SCORM test resource files (a valid minimal SCORM zip fixture). Add one to `src/test/resources/` similar to how `uploadAPK.apk` exists for the APK manager tests.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/factory/MimeTypeManagerFactory.scala
```diff
@@ -23,6 +23,7 @@ object MimeTypeManagerFactory {
 			"application/vnd.ekstep.content-collection" -> new CollectionMimeTypeMgrImpl,
 			"application/vnd.ekstep.plugin-archive" -> new PluginMimeTypeMgrImpl,
 			"application/vnd.ekstep.h5p-archive" -> new H5PMimeTypeMgrImpl,
```
sntiwari1, line 25: **Missing test: MimeTypeManagerFactoryTest needs a SCORM case**

`MimeTypeManagerFactoryTest.scala` has a dedicated test for every mime type registered in the factory (H5P, HTML, ECML, Collection, Plugin, APK, Document, YouTube, Asset). This PR adds `application/vnd.ekstep.scorm-archive` to the factory but does not add a corresponding test to verify that `getManager("Resource", "application/vnd.ekstep.scorm-archive")` returns an instance of `ScormMimeTypeMgrImpl`.

Please add:
```scala
"getManager with mimeType application/vnd.ekstep.scorm-archive" should "give instance of ScormMimeTypeMgrImpl" in {
  val mgr = MimeTypeManagerFactory.getManager("Resource", "application/vnd.ekstep.scorm-archive")
  assert(null != mgr)
  assert(mgr.isInstanceOf[ScormMimeTypeMgrImpl])
}
```

platform-modules/mimetype-manager/src/test/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImplTest.scala
```diff
@@ -0,0 +1,37 @@
+package org.sunbird.mimetype.mgr.impl
+
+import com.google.common.io.Resources
```
sntiwari1: **Nit: Unused imports**

`com.google.common.io.Resources` (line 3) and `scala.concurrent.Future` (line 13) are imported but never referenced in the test file. These should be removed to keep the file clean. `Resources` would only be needed when loading test resource files (a valid SCORM zip fixture), which are not present yet.

sntiwari1 (changes requested):
## Code Review: SCORM Package Ingestion Support

Thank you for adding SCORM support to the platform. The overall structure — a new `ScormMimeTypeMgrImpl` following the existing H5P/HTML manager pattern, registration in `MimeTypeManagerFactory`, and schema addition — is the correct approach. The launch file integrity check (verifying the `imsmanifest.xml` href exists in the package) is a welcome addition.

However, there are **three security issues** (two of them blocking), a resource leak, a performance concern, and significant gaps in test coverage that must be addressed before this can be merged.

---

### Blocking Issues

**1. XXE (XML External Entity) Injection** — [inline comment on line 24]
`scala.xml.XML.loadFile` enables external entity resolution by default. A malicious `imsmanifest.xml` inside an uploaded zip can exfiltrate local files or trigger SSRF. Must be fixed with a hardened SAX parser that disables DOCTYPE and external entity features.

**2. Launch file path traversal** — [inline comment on line 30]
The `href` attribute from the manifest is used unsanitized to construct a filesystem path. An attacker can supply `"../../../etc/passwd"` or an absolute path (`"/etc/passwd"`) to bypass or misuse the existence check. The resolved path must be canonicalized and confirmed to be inside `extractionBasePath` before use.

**3. Zip Slip vulnerability** — [inline comment on line 23]
The shared `BaseMimeTypeManager.extractPackage` does not validate zip entry paths before writing. This affects all archive managers, but SCORM adds another entry point. Path normalization and boundary checking must be added before or inside `extractPackage`.

---

### Non-Blocking Issues (should fix before merge)

**4. Temp directory resource leak** — [inline comment on line 22]
The extracted `extractionBasePath` directory is never deleted — not on the happy path, not on exception. Use a `try/finally` block with the existing `delete()` helper. This will cause disk exhaustion in production under any upload load.

**5. Missing test coverage** — [inline comment on test line 19]
Only one test exists (null file guard). The happy path, the missing-manifest error, the missing-launch-file error (the primary feature of this PR), the URL upload overload, review, and publish are all untested. A minimal SCORM zip fixture needs to be added to `src/test/resources/`.

**6. Missing factory test** — [inline comment on MimeTypeManagerFactory.scala]
`MimeTypeManagerFactoryTest` tests every other mime type mapping. The SCORM entry must be added to maintain that pattern.

**7. Blocking I/O on dispatcher thread** — [inline comment on line 36]
`uploadArtifactToCloud`, `extractPackage`, and XML parsing all run synchronously before the `Future {}` block is constructed. This is consistent with existing managers but should be wrapped with `scala.concurrent.blocking` at minimum.

---

### Minor

**8. Unused imports in test** — [inline comment on test line 3]
`com.google.common.io.Resources` and `scala.concurrent.Future` are imported but unused.

**9. Indentation inconsistency**
The new `ScormMimeTypeMgrImpl` uses 4-space indentation. All other files in this package (H5P, HTML, APK, etc.) use tab indentation. Please match the existing style.

**10. `getOrElse("index.html")` silent fallback**
Silently defaulting the launch file to `"index.html"` when no `resource` with an `href` is found in the manifest masks malformed packages. SCORM 1.2 requires a launch resource; a missing one should raise a `ClientException` for strict validation.

---

### What is Good

- Correct class hierarchy: extends `BaseMimeTypeManager` and mixes in `MimeTypeManager`, matching all other managers exactly.
- Namespace-agnostic XML traversal with `\\ "_"` + label match is the right approach for IMS manifest files that may use namespace prefixes.
- Schema addition is minimal and correct.
- The three-overload structure (file upload, URL upload, review, publish) is complete and consistent with the trait contract.
- Error codes and messages follow the existing platform conventions (`ERR_INVALID_FILE`, `ClientException`).

Please resolve items 1–6 (security, resource leak, and test gaps) before re-requesting review.

sntiwari1 (changes requested):
## Code Review: SCORM Package Ingestion — Post-Fix Review

Good progress — the five issues from the original review have been addressed, with one critical exception that is blocking merge. See details below.

### ✅ Fixed from prior review

| # | Issue | Fix |
|---|---|---|
| 1 | XXE injection via `scala.xml.XML.loadFile` | ✅ `getSecureXml()` correctly disables DOCTYPE and external entity features |
| 2 | Launch file path traversal | ✅ Canonical path comparison (`canonicalLaunch.startsWith(canonicalBase + File.separator)`) |
| 3 | Temp directory resource leak | ✅ `try/finally { delete(new File(extractionBasePath)) }` |
| 4 | Blocking I/O on dispatcher thread | ✅ `blocking { }` wrappers applied throughout |
| 5 | Factory test missing | ✅ `MimeTypeManagerFactoryTest` now has the SCORM case |

---

### ❌ Blocking: Zip Slip fix is incomplete

The patch added to `BaseMimeTypeManager.extractPackage` uses `toAbsolutePath()` but does **not** call `normalize()`. This means path components like `..` are preserved in the result, and the `startsWith` check can be bypassed:

```
baseDirPath = /tmp/abc/xyz         (absolute, no normalize)
targetPath  = /tmp/abc/xyz/../evil.txt  (absolute, but un-normalized)

targetPath.startsWith(baseDirPath) == true  ← attacker bypasses the check
```

The fix is a one-word change:

```scala
// Current (vulnerable):
val baseDirPath = Paths.get(basePath).toAbsolutePath
val targetPath  = Paths.get(basePath + File.separator + entry.getName).toAbsolutePath

// Fixed:
val baseDirPath = Paths.get(basePath).toAbsolutePath.normalize()
val targetPath  = Paths.get(basePath + File.separator + entry.getName).toAbsolutePath.normalize()
```

`normalize()` collapses `..` and `.` components, making `startsWith` reliable. This must be fixed before merge since `extractPackage` is shared by all archive mime-type managers (ECML, HTML, H5P, Plugin, and now SCORM).

---

### ❌ Blocking: `ScormMimeTypeMgrImplTest.scala` is still empty

The file is a 0-byte placeholder (`sha: e69de29bb2d1d6434b8b29ae775ad8c2e48c5391`). The prior review requested six test scenarios; none were added. The tests requested are not nice-to-have — the two security fixes (path traversal + launch file validation) added in this PR have zero test coverage. Please add at minimum:

1. **Happy path** — valid SCORM zip with correct `imsmanifest.xml` and existing launch file succeeds
2. **Missing manifest** — zip without `imsmanifest.xml` throws `ClientException("ERR_INVALID_FILE", ...)`
3. **Missing launch file** — manifest references a file absent from the zip throws `ClientException("ERR_INVALID_FILE", ...)`
4. **Path traversal in manifest** — manifest href `"../../../etc/passwd"` throws `ClientException("ERR_INVALID_FILE", "Invalid launch file path!")`
5. **`review` with no artifact** — `validate()` throws `ClientException("VALIDATOR_ERROR", ...)`

Tests 2–4 directly exercise the new security logic. Without them, a regression in `getValidatedLaunchFile` or the path traversal check would go undetected.

---

### Non-blocking notes

- The `launchFile` value stored in node metadata is not validated against `null` or empty string. An `href=""` in the manifest would pass both the canonical-path check (it resolves to `extractionBasePath` itself, which exists as a directory) and the `exists()` check, then persist an empty string. A simple `if (launchFile.isEmpty) throw new ClientException(...)` guard before the path check would close this.
- Minor: mixed indentation between the outer `upload` method (tabs) and inner private methods (4-space). Not a blocker but worth cleaning for consistency with the rest of the file.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/BaseMimeTypeManager.scala
```diff
@@ -135,12 +135,19 @@ class BaseMimeTypeManager(implicit ss: StorageService) {
 
 	def extractPackage(file: File, basePath: String) = {
 		val zipFile = new ZipFile(file)
+		val baseDirPath = Paths.get(basePath).toAbsolutePath
 		for (entry <- zipFile.entries().asScala) {
```
sntiwari1, line 139: **[Blocking] Zip Slip fix is incomplete — `toAbsolutePath` without `normalize()` is still bypassable**

`toAbsolutePath()` makes the path absolute but does **not** collapse `..` components. A zip entry named `../../evil.txt` produces:

```
baseDirPath = /tmp/abc/xyz          ← toAbsolutePath, no normalize
targetPath  = /tmp/abc/xyz/../../evil.txt  ← toAbsolutePath, no normalize
```

`targetPath.startsWith(baseDirPath)` evaluates to `true` here because `Path.startsWith` does a component-by-component prefix match — `/tmp`, `abc`, `xyz` match the first three components of `targetPath` before it hits `..`, so the traversal is never detected.

Change both lines to add `.normalize()`:

```scala
val baseDirPath = Paths.get(basePath).toAbsolutePath.normalize()
// ...
val targetPath = Paths.get(basePath + File.separator + entry.getName).toAbsolutePath.normalize()
```

`normalize()` resolves `..` and `.` components before the `startsWith` check runs, making it reliable. This affects all archive mime-type managers that call `extractPackage` (ECML, HTML, H5P, Plugin, SCORM).

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,119 @@
...
+        val launchFile = ref.flatMap { r =>
+            (xml \\ "resource").find(n => (n \@ "identifier") == r).map(_ \@ "href")
+        }.getOrElse {
+            TelemetryManager.error("ERR_INVALID_FILE:: Launch file not found in imsmanifest.xml")
+            throw new ClientException("ERR_INVALID_FILE", "Launch file not found in imsmanifest.xml!")
+        }
+
+        // Validate launchFile containment and existence
+        val combinedFile = new File(extractionBasePath, launchFile)
+        val canonicalBase = new File(extractionBasePath).getCanonicalPath
```
sntiwari1, line 78: **[Non-blocking] Empty `href` attribute passes both security checks and persists as the launch file**

If the matched `<resource>` element has an empty `href` attribute (e.g. `href=""`), `launchFile` will be `""`. The canonical path check and `exists()` check both pass silently:

```
combinedFile     = new File("/tmp/abc/xyz", "")  →  resolves to /tmp/abc/xyz  (the extraction dir itself)
canonicalLaunch  = /tmp/abc/xyz
canonicalBase    = /tmp/abc/xyz

canonicalLaunch.startsWith(canonicalBase + "/")  →  false!  — wait, the dir check would FAIL here
```

Actually on re-examination: `canonicalLaunch == canonicalBase`, so `startsWith(canonicalBase + File.separator)` is `false` — the path traversal check *does* reject it. However, an `href` of `"./"` or `"subdir/"` (a directory, not a file) would produce a path that passes the containment check *and* `exists()` returns true for the directory, letting an invalid launch file through.

Add an explicit non-empty, non-directory guard before the path checks:

```scala
if (launchFile.isEmpty)
  throw new ClientException("ERR_INVALID_FILE", "Launch file href is empty in imsmanifest.xml!")

// ... existing canonical path check ...

if (!combinedFile.exists() || combinedFile.isDirectory)
  throw new ClientException("ERR_INVALID_FILE", "The launch file '" + launchFile + "' is missing or is not a file!")
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,119 @@
...
+
+import scala.concurrent.{ExecutionContext, Future, blocking}
+import scala.xml.Elem
+import scala.xml.factory.XMLLoader
+
+class ScormMimeTypeMgrImpl(implicit ss: StorageService) extends BaseMimeTypeManager()(ss) with MimeTypeManager {
+
+    override def upload(objectId: String, node: Node, uploadFile: File, filePath: Option[String], params: UploadParams)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
+        validateUploadRequest(objectId, node, uploadFile)
+        TelemetryManager.info("SCORM content upload for objectId:: " + objectId)
```
sntiwari1, line 21: **[Blocking] `startsWith` on un-normalized paths is bypassable — same issue as in `extractPackage`**

`getValidatedLaunchFile` uses `getCanonicalPath()` for the path traversal check, which is correct and safe. However, `extractPackage` (called just before this on line 28) still uses `toAbsolutePath` without `normalize()`, so a malicious zip entry can write a file outside `extractionBasePath` *before* this method ever runs. Fix `extractPackage` first (see comment on `BaseMimeTypeManager.scala` line 138), and the protection here will be sound.

Additionally, `ScormMimeTypeMgrImplTest.scala` is currently an empty file (0 bytes). None of the logic in `getValidatedLaunchFile` — including this path traversal check — has any test coverage. Please add at minimum:

```scala
// Test 1: missing imsmanifest.xml
"upload" should "throw ClientException for zip missing imsmanifest.xml" in {
  val file = createZipWithoutManifest()
  val ex = intercept[ClientException] { Await.result(scormMgr.upload("do_1", getNode(), file, None, UploadParams()), 10.seconds) }
  ex.getErrCode shouldBe "ERR_INVALID_FILE"
}

// Test 2: manifest present but referenced launch file absent
"upload" should "throw ClientException when launch file is absent from package" in { ... }

// Test 3: manifest with path-traversal href
"getValidatedLaunchFile" should "throw ClientException for href with path traversal" in { ... }
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,119 @@
...
+        launchFile
+    }
+
+    private def getSecureXml(manifestFile: File): Elem = {
+        val spf = SAXParserFactory.newInstance()
+        spf.setNamespaceAware(true)
+        spf.setFeature("http://xml.org/sax/features/external-general-entities", false)
+        spf.setFeature("http://xml.org/sax/features/external-parameter-entities", false)
+        spf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true)
+
```
sntiwari1, line 101: **[Nit] Consider adding `load-external-dtd` disable for defence-in-depth**

`disallow-doctype-decl` already prevents DOCTYPE declarations, which blocks the most common XXE vectors. For additional hardening, some environments recommend also setting:

```scala
spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false)
```

This is not required given `disallow-doctype-decl` is already present, but it makes the intent explicit and defends against edge cases in non-Apache SAX implementations where `disallow-doctype-decl` may not be recognised.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,119 @@
...
+    
+    override def upload(objectId: String, node: Node, fileUrl: String, filePath: Option[String], params: UploadParams)(implicit ec: ExecutionContext): Future[Map[String, AnyRef]] = {
+        validateUploadRequest(objectId, node, fileUrl)
+        val file = copyURLToFile(objectId, fileUrl)
+        upload(objectId, node, file, filePath, params)
+    }
+
+    override def review(objectId: String, node: Node)(implicit ec: ExecutionContext, ontologyEngineContext: OntologyEngineContext): Future[Map[String, AnyRef]] = {
+        validate(node, "| [SCORM file should be uploaded for further processing!]")
+        Future(getEnrichedMetadata(node.getMetadata.getOrDefault("status", "").asInstanceOf[String]))
```
sntiwari1, line 117: **[Nit] Validation message starts with a stray `|` character**

The message `"| [SCORM file should be uploaded for further processing!]"` has a leading pipe. Compare with `H5PMimeTypeMgrImpl`:

```scala
// H5PMimeTypeMgrImpl
validate(node, "| [H5P file should be uploaded for further processing!]")
```

If this is the established convention in this codebase, ignore this comment. But if the `|` is unintentional, the message should be:

```scala
validate(node, "[SCORM file should be uploaded for further processing!]")
```

Same applies to the `publish` override on line 117.

README.md
```diff
@@ -13,8 +13,9 @@ Play Framework APIs for the Sunbird Knowledge Platform. Each service exposes RES
    - [Step 2 — Start infrastructure](#step-2--start-infrastructure)
    - [Step 3 — Initialize YugabyteDB keyspaces](#step-3--initialize-yugabytedb-keyspaces)
    - [Step 4 — Initialize Elasticsearch indices](#step-4--initialize-elasticsearch-indices)
-   - [Step 5 — Build the project](#step-5--build-the-project)
-   - [Step 6 — Run a service](#step-6--run-a-service)
+   - [Step 5 — Seed data](#step-5--seed-data)
```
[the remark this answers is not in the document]
romitshah02: resolved

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/BaseMimeTypeManager.scala
romitshah02: I have updated extractPackage in BaseMimeTypeManager.scala to apply .normalize() to both baseDirPath and targetPath as requested. This ensures that path components like .. are collapsed before the startsWith check, making the containment validation robust across all archive mime-type managers.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
romitshah02: I've updated the validation logic in getValidatedLaunchFile to include an explicit guard for empty strings and to verify that the resolved path exists and is a file, not a directory (if (!combinedFile.exists() || combinedFile.isDirectory)). This prevents invalid launch file configurations from being persisted.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
romitshah02: I have addressed the Zip Slip vulnerability via path normalization in BaseMimeTypeManager.scala and added the requested test coverage to ScormMimeTypeMgrImplTest.scala.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
romitshah02: I have added the load-external-dtd feature flag to getSecureXml as requested to further harden the XML parsing against XXE edge cases.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
romitshah02: I have removed the unintentional leading pipe character from the validation messages in both the review and publish methods.

romitshah02:
Implemented comprehensive test coverage for ScormMimeTypeMgrImpl in ScormMimeTypeMgrImplTest.scala. These tests cover valid SCORM packages, missing imsmanifest.xml files, missing launch files, and path traversal attempts, confirming that the security logic behaves as expected.

PR #1238, opened 2026-05-27
feat: semantic search with hybrid RRF fusion and content enrich API

chethann007:
# Summary

Introduces semantic search (kNN vector similarity) to the search API, extends it with hybrid text+semantic fusion, and adds an enrich API endpoint for triggering async metadata enrichment without republishing.

## Breaking changes

- `compositesearch.index.name` default corrected from `kp_audit_log` to `compositesearch`. Any deployment overriding this to `kp_audit_log` intentionally should keep that override.

## Semantic search

A new `search_mode=semantic` runs kNN queries against pre-indexed vector embeddings, ranking results by cosine similarity. Filters from the request are correctly inherited into the kNN filter clause — fixes ship for two silent correctness bugs: full-text `propertyName='*'` leaking into filters (dropping all docs) and fuzzy mode bypassing the `BoolQueryBuilder` filter set. Semantic results always surface cosine score; default name/date sort is skipped so kNN ranking is preserved. `min_score` threshold supported to drop low-confidence matches.

## Hybrid search

`search_mode=hybrid` runs text and semantic queries in parallel and fuses results via Reciprocal Rank Fusion — `score = Σ 1/(k + rank_i)`, k=60. Each leg runs in a fresh `SearchProcessor` instance to avoid race conditions on mutable sort state. Results merge into a single ranked list with `score_components` (text_rank, semantic_rank) per hit. Pagination applied post-fusion; facets borrowed from text leg. Existing text-mode clients unaffected.

## Enrich API

`POST /v3/enrich` (search-service) accepts a list of content identifiers, validates each exists, derives objectType from mimeType, and emits `BE_JOB_REQUEST` Kafka events with `action=enrich` to the publish topic — triggering `EnrichOnlyFunction` in knowlg-publish to re-emit enriched metadata without running the full publish flow. Useful for bulk backfilling existing content into the semantic index after field config or framework changes.

## Test plan
- [ ] `search_mode=semantic` with filters — verify filters applied correctly, cosine scores present, kNN ordering preserved
- [ ] `min_score` — verify hits below threshold excluded
- [ ] `POST /v3/enrich` (search-service) — verify Kafka event emitted, EnrichOnlyFunction picks it up
- [ ] Existing `search_mode=text` requests — verify no regression


content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
```diff
@@ -0,0 +1,89 @@
...
+import org.sunbird.kafka.client.KafkaClient
+import org.sunbird.telemetry.logger.TelemetryManager
+
+import java.util
+import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters._
+
+object EnrichManager {
+
+  private val kfClient = new KafkaClient
```
sntiwari1: **[Bug / Resource Leak]** \`private val kfClient = new KafkaClient\` is a module-level singleton constructed at class-load time with no lifecycle management.

1. **Not closed**: The client (and its underlying Kafka producer/connections) is never shut down. If the JVM shut-down hook does not flush the producer, in-flight messages can be lost.
2. **Thread safety**: `KafkaClient.send()` is called from multiple concurrent `Future` completions (one per identifier in the batch). Verify that `KafkaClient` is thread-safe before relying on this sharing. If it is not, wrap with a `synchronized` block or use a per-call instance.
3. **Better pattern**: inject the client via `EnrichManager.triggerEnrich(request, kfClient)(...)` or resolve it from a singleton provided by the DI graph (same pattern the publish actors use).

content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
```diff
@@ -0,0 +1,89 @@
...
+      val readReq = new Request(request)
+      readReq.put(ContentConstants.IDENTIFIER, id)
+      readReq.put("fields", new util.ArrayList[String]())
+      readReq.getContext.put("objectType", "Content")
+      readReq.getContext.put("schemaName", "content")
+      DataNode.read(readReq).map { node =>
+        val mimeType   = Option(node.getMetadata.get("mimeType")).map(_.toString).getOrElse("")
+        val objectType = resolveObjectType(mimeType)
+        id -> Some((objectType, mimeType))
+      }.recover { case _ => id -> None }
```
sntiwari1: **[Bug - Silent Error Swallow]** `recover { case _ => id -> None }` catches **every** Throwable and maps it to `None`, which makes the identifier appear "not found". This hides programming errors (NPE, serialization failures, unexpected graph exceptions) that should propagate as 500s.

Prefer a typed recover that only handles expected not-found cases:
```scala
.recover {
  case e: ResourceNotFoundException => id -> None
  case e: Exception =>
    TelemetryManager.error(s"Failed to read node $id: ${e.getMessage}", e)
    id -> None  // or re-throw if you want a hard failure
}
```
At minimum log the exception so ops can distinguish "content missing" from "graph unreachable".

content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
```diff
@@ -0,0 +1,89 @@
...
+        val ets = System.currentTimeMillis()
+        val mid = s"LP.$ets.${java.util.UUID.randomUUID()}"
+        val event = Map(
+          "eid"     -> "BE_JOB_REQUEST",
+          "ets"     -> ets,
+          "mid"     -> mid,
+          "actor"   -> Map("id" -> "content-enrich-api", "type" -> "System"),
+          "context" -> Map("pdata" -> Map("ver" -> "1.0", "id" -> "org.sunbird.platform")),
+          "object"  -> Map("id" -> id, "ver" -> "1.0"),
+          "edata"   -> Map(
```
sntiwari1: **[Error handling gap]** `kfClient.send(...)` is fire-and-forget inside a `foreach`. If the Kafka send throws (broker unreachable, serialization error), the exception propagates out of `foreach` and the Future fails **after** some events were already emitted for earlier identifiers — partial emission with no way for the caller to know which IDs were actually enqueued.

Consider wrapping the send in a `Try` and accumulating per-ID failures, or returning a partial-success shape in the response body.

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
```diff
@@ -0,0 +1,212 @@
...
+        EmbeddingClient client = EmbeddingClientFactory.get();
+        float[] vec = cache.get(client.getName(), client.getVersion(), query);
+        if (vec == null) {
+            vec = client.embed(query);
+            cache.put(client.getName(), client.getVersion(), query, vec);
+        }
+
+        QuantizationStrategy quantizer = QuantizationStrategyFactory.get(
+                getString(dto.getSemanticParams(), "quantization_strategy",
+                        getString("semantic_search.quantization_strategy", "int8")));
```
sntiwari1, line 90: **[Critical - Actor thread blocking]** `client.embed(query)` is a synchronous HTTP call (up to `timeout` seconds, default 5s) happening **on the Pekko actor dispatcher thread**. This blocks the entire actor and starves other messages.

`DESIGN.md` explicitly says:
> Embedding HTTP must not run on the Akka actor dispatcher. Dedicated ForkJoinPool (parallelism = Math.max(8, cpu*2)) named `semantic-embedding-pool`.

`build()` currently returns a `QueryBuilder` (a synchronous value), which makes it structurally impossible to delegate the embedding call to another thread pool without changing the interface. Phase 2b of the implementation plan should change `QueryStrategy.build()` to return a `CompletableFuture<QueryBuilder>` (or a Scala `Future`) so the embedding call can be offloaded. Alternatively, execute the embedding call before entering `processSearchQuery` and pass the pre-computed vector into the DTO.

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
```diff
@@ -0,0 +1,212 @@
...
+            if (existing instanceof BoolQueryBuilder) {
+                BoolQueryBuilder bx = (BoolQueryBuilder) existing;
+                for (QueryBuilder f : bx.filter())  finalBool.filter(f);
+                for (QueryBuilder m : bx.mustNot()) finalBool.mustNot(m);
+                // Every remaining must is a property filter (full-text leg
+                // already excluded). Demote to filter so kNN owns scoring.
+                for (QueryBuilder m : bx.must())   finalBool.filter(m);
+                // Implicit-filter shoulds inherited from getSearchQuery — keep
+                // as filters so they constrain rather than just boost.
+                for (QueryBuilder s : bx.should()) finalBool.filter(s);
```
sntiwari1, line 188: **[Comment / code mismatch]** The comment on line ~133 says:
```
//  - should() → dropped entirely
```
But this line promotes every `should()` clause to a hard filter:
```java
for (QueryBuilder s : bx.should()) finalBool.filter(s);
```
Implicit-filter shoulds (`visibility`, `status` from `getImplicitFilterProperties`) must constrain results, so promoting them to `filter` is correct. But if the text query also injected soft-constraint shoulds (from `SoftConstraintMode`), those would incorrectly become hard filters here and exclude documents that merely do not match the boost.

Either update the comment to reflect actual behaviour, or strip soft-constraint shoulds before promoting: check the source of each `should` clause and only promote implicit-filter ones.

search-api/search-core/src/main/java/org/sunbird/search/processor/HybridSearchExecutor.java
```diff
@@ -0,0 +1,173 @@
...
+                                            int rrfK,
+                                            int limit,
+                                            int offset) {
+        List<Map<String, Object>> textHits = (List<Map<String, Object>>) textResp.getOrDefault("results", Collections.emptyList());
+        List<Map<String, Object>> semHits  = (List<Map<String, Object>>) semResp.getOrDefault("results", Collections.emptyList());
+
+        List<List<Map<String, Object>>> sources = new ArrayList<>(2);
+        sources.add(textHits);
+        sources.add(semHits);
+
```
sntiwari1, line 104: **[Performance / design]** `ExecutionContext.Implicits$.MODULE$.global()` is Scala's global fork-join pool, which is the same pool the Pekko dispatcher uses under default config. The `.zip(semF)` continuation runs on this pool, meaning it competes with actor message processing.

`DESIGN.md` specifies a dedicated `semantic-embedding-pool` for exactly this reason. Wire that pool here (as a `java.util.concurrent.Executor` passed in or looked up from a named dispatcher) so the fusion step does not borrow from the dispatcher.

search-api/search-core/src/main/java/org/sunbird/search/processor/HybridSearchExecutor.java
```diff
@@ -0,0 +1,173 @@
...
+                ExecutionContext.Implicits$.MODULE$.global()
+        );
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Map<String, Object> fuse(Map<String, Object> textResp,
+                                            Map<String, Object> semResp,
+                                            int rrfK,
+                                            int limit,
+                                            int offset) {
```
sntiwari1, line 97: **[Missing fallback for hybrid mode]** `textF.zip(semF)` fails the whole request if the semantic leg fails (embedding timeout, circuit open, etc.). `DESIGN.md` and `API_SPEC.md` specify that `params.degraded = true` should be returned and results should fall back to text-only.

Consider using `semF.recover(...)` before the zip — on failure substitute an empty results map — so a semantic outage does not surface as a 500 on hybrid requests.

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingCache.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public void clear() {
+        synchronized (store) { store.clear(); }
+    }
+
+    private static String key(String service, String model, String text) {
+        try {
+            MessageDigest md = MessageDigest.getInstance("SHA-256");
+            String raw = service + ":" + model + ":" + text;
+            byte[] digest = md.digest(raw.getBytes(java.nio.charset.StandardCharsets.UTF_8));
```
sntiwari1, line 64: **[Cache correctness]** The cache key is `service + ":" + model + ":" + text`, but the callers pass `client.getName()` ("openai") and `client.getVersion()` ("1.0") for `service` and `model`. Neither of those encodes the actual model name (e.g. `text-embedding-3-small`).

If the deployment is switched from `text-embedding-3-small` to `text-embedding-ada-002` (different vector space!) without restarting the service, the cache will return stale `3-small` vectors keyed under `"openai:1.0:..."` for the same queries, silently corrupting semantic recall.

Fix: include the actual model name in the key. Pass `client.getVersion()` as-is but derive a stable unique key from the full config (e.g. `serviceName + ":" + modelName + ":" + text` where `modelName` comes from `EmbeddingClientConfig.getModel()`).

search-api/search-core/src/main/java/org/sunbird/search/strategy/QueryStrategyFactory.java
```diff
@@ -0,0 +1,74 @@
...
+                        "search_mode=" + key + " is not enabled on this server");
+            }
+            throw new ClientException(
+                    SearchConstants.ERR_INVALID_SEARCH_MODE,
+                    "Unsupported search_mode: " + mode);
+        }
+        return strategy;
+    }
+
+    /** Test-only. Do not call from production code paths. */
```
sntiwari1, line 74: **[Missing circuit breaker / degraded path]** `DESIGN.md` and the API spec describe a circuit breaker: after N consecutive embedding failures the breaker opens, and subsequent semantic/hybrid requests degrade gracefully to text mode with `params.degraded=true` rather than returning a 500. That class (`EmbeddingCircuitBreaker`) is referenced in the FLOWCHART but is absent from this PR.

Without it, a sustained embedding-service outage causes every `search_mode=semantic` request to throw a `RuntimeException` from `SemanticQueryStrategy.build()`, which propagates as a 500 to clients.

At minimum, add a `TODO` here with a ticket reference so this gap is tracked, or gate the PR on the circuit breaker being present.

sntiwari1 (commented):
## Code Review — PR #1238: Semantic Search + Enrich API

### Summary

Well-scoped and thoroughly documented PR. The strategy refactor is clean, the RRF fusion is a correct pure function, quantization mirrors the job faithfully, and the docs (DESIGN.md, FLOWCHART.md, OpenAPI spec) are unusually thorough. The Enrich API is purposeful.

There are two critical issues that should block merge, plus several correctness bugs and design gaps detailed in the inline comments.

---

### Critical (block merge)

1. **Actor thread blocking** — `SemanticQueryStrategy.java:68`: `client.embed()` is a synchronous HTTP call (up to 5s) running on the Pekko dispatcher. DESIGN.md explicitly says it must not. The `QueryStrategy.build()` signature returning a synchronous `QueryBuilder` needs to change (to a `CompletableFuture`) or the embed call must be done before entering `processSearchQuery`.

2. **No circuit breaker / degraded path** — `QueryStrategyFactory.java:69`: `EmbeddingCircuitBreaker` is described in DESIGN.md and FLOWCHART.md but not present. An embedding outage causes a 500 on every semantic/hybrid request. The `params.degraded` response contract from API_SPEC.md is unimplemented.

---

### Correctness bugs

3. **Cache key missing model name** — `EmbeddingCache.java:64`: Key is `client.getName()+":"+client.getVersion()` which is always `"openai:1.0"`. Swapping from `text-embedding-3-small` to `ada-002` hot-reuses stale vectors from an incompatible space, silently corrupting recall.

4. **Comment / code mismatch on `should()`** — `SemanticQueryStrategy.java:157`: Comment says "should() → dropped entirely" but the code promotes all shoulds to hard filters. Soft-constraint shoulds then incorrectly exclude documents that merely lack the boost.

5. **Silent error swallow** — `EnrichManager.scala:41`: `recover { case _ => id -> None }` treats every Throwable (graph timeout, NPE, serialization error) as "content not found", turning a 500 into a silent 400.

---

### Design gaps

6. **Hybrid mode has no fallback** — `HybridSearchExecutor.java:80`: `textF.zip(semF)` propagates a semantic failure as a 500. Should `recover` the semantic leg and degrade to text-only with `params.degraded=true`.

7. **Wrong execution context** — `HybridSearchExecutor.java:87`: Uses Scala global EC (same pool as Pekko dispatcher). DESIGN.md specifies a dedicated `semantic-embedding-pool`.

8. **Kafka client lifecycle** — `EnrichManager.scala:18`: Module-level singleton never closed; verify `KafkaClient.send()` is thread-safe before relying on concurrent use.

9. **Partial Kafka emission** — `EnrichManager.scala:62`: A mid-batch send failure leaves some events emitted and some not, with no partial-success signal to the caller.

---

### What's well done

- **RRF** (`RrfFusion.java`) correct and cleanly generic; `LinkedHashMap` tie-breaking is a nice touch.
- **Int8 quantization** faithfully mirrors the job's normalized/unnormalized branch.
- **Filter inheritance** in `SemanticQueryStrategy` — the two-pass approach correctly avoids the `FunctionScoreQueryBuilder` wrapping problem; the long inline comment explains it well.
- **Hybrid leg isolation** — fresh `SearchProcessor` per leg avoids the `relevanceSort` race. Good catch.
- **`enabled = false` default** — text clients are fully unaffected at deploy time.
- **E5 SSRF guard** — private-range + cluster-local hostname validation is the right call.
- **Documentation** — DESIGN.md, FLOWCHART.md, API_SPEC.md, and OpenAPI are production-quality.

---

### Minor nits

- `ApiId.TRIGGER_ENRICH` is `val` while surrounding constants use `final val` — minor consistency.
- No unit tests for any new Java classes in this PR. The implementation plan defers them to Phase 2a, but they should land with or before the code that depends on them.
- `SearchConstants.search_mode` and `SearchConstants.semantic` use lowercase names; surrounding constants are uppercase.

content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
chethann007: Lifecycle: KafkaClient constructor calls registerShutdownHook() (line 18) which registers a JVM shutdown hook that calls producer.flush() + producer.close(). In-flight messages are flushed on JVM exit. No leak.

Thread safety: KafkaProducer.send() is explicitly documented thread-safe by Kafka (producer is designed for concurrent use across threads). No sync block needed.

Pattern: private val kfClient = new KafkaClient is the established codebase convention — identical in RetireManager:31, PublishManager:21, DIALManager:34, UploadManager:28. Injecting via parameter would be inconsistent with every other manager and adds no correctness benefit since the producer is thread-safe and lifecycle-managed.

content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
chethann007: Applied the changes, thanks!

content-api/content-actors/src/main/scala/org/sunbird/content/util/EnrichManager.scala
chethann007: Already fixed — commit e81a9c6b. Each send wrapped in Try, response now returns succeeded/failed arrays. Partial emission visible to caller.

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
chethann007: Applied the changes, thanks!

search-api/search-core/src/main/java/org/sunbird/search/processor/HybridSearchExecutor.java
chethann007: global() ≠ Pekko dispatcher — separate pools. fuse() is pure CPU, microseconds — wrong tool to isolate. SemanticEmbeddingPool is for blocking I/O only.

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
chethann007: Applied the changes. thanks!

search-api/search-core/src/main/java/org/sunbird/search/processor/HybridSearchExecutor.java
chethann007: Applied the changes, thanks for the suggestion!

search-api/search-core/src/main/java/org/sunbird/search/strategy/QueryStrategyFactory.java
chethann007: Applied the changes, thanks for the suggestion!

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingCache.java
chethann007: Applied the changes, thanks for the suggestion!

COMMIT_RANGE_ANALYSIS.md
```diff
@@ -0,0 +1,381 @@
+# Commit Range Analysis: f3ab89d to 60e28be (inclusive)
```
sntiwari1: This analysis document should not be committed to the repository. It is an internal development artifact and adds noise to `git log`, `git blame`, and search results. Please remove this file before merge (move it to a local notes file or a Confluence/Notion page if you'd like to keep the analysis).

sntiwari1 (commented):
## Code Review — PR #1238: feat: semantic search with hybrid RRF fusion and content enrich API

### Summary

This is a well-designed, comprehensive feature. The strategy pattern, pluggable embedding stack, int8 quantization, RRF fusion, and semantic+hybrid modes are all solid. The edge-case fixes (fuzzy save/restore, filter-leg strip, fresh SearchProcessor per hybrid leg) show careful attention. Test coverage for the pure-function components (RrfFusion, Int8Quantization, EmbeddingCache) is good.

Several issues need fixing before merge — two are high-severity (mutable cache reference returned, enabled=true in config), and a few more are medium severity (blocking I/O dispatcher, per-request ES init, no batch-size cap). Inline comments below call each one out at the exact location.

---

### High-Severity Issues

1. **`EmbeddingCache.get()` returns mutable array** — cache can be silently corrupted if caller writes to the returned float[]. Defensive copy required.
2. **`semantic_search.enabled = true` shipped in application.conf** — the PR description says this should default to `false`. This turns on semantic search for all deployments that pick up this config.
3. **`compositesearch.index.name` changed** — flipped from `kp_audit_log` to `compositesearch` in search-service config; this is a silent breaking change for operators who rely on the default and deserves its own entry in the PR description.

### Medium-Severity Issues

4. **`EnrichActor` does blocking I/O on Scala global EC** — ES multi-get + Kafka sends run on `ExecutionContext.Implicits$.MODULE$.global()`, which is the ForkJoinPool. Blocking calls starve it. Use a dedicated I/O dispatcher (or mirror the SemanticEmbeddingPool pattern already in this PR).
5. **`ElasticSearchUtil.initialiseESClient()` called per request in EnrichActor** — expensive; should happen once at actor startup.
6. **No max identifiers cap in EnrichActor** — unbounded list accepted; a 10,000-identifier request does one ES multi-get and loops Kafka sends in the same thread. Add a configurable cap (e.g. 100).
7. **`SemanticQueryStrategy` promotes `should` clauses to `filter`** — the implicit-filter `should` wrap is promoted assuming it is visibility-constraining, but a `should`-only BoolQuery in filter context applies implicit `minimum_should_match=1`, which IS intended. The risk is any future `should` clause that is a scoring boost rather than a visibility gate being silently hardened into a filter. The assumption needs a comment or an assertion.
8. **`COMMIT_RANGE_ANALYSIS.md` should not be committed** — internal analysis doc adds no value to the repo history and should be removed before merge.

### Low-Severity / Suggestions

9. **`SemanticEmbeddingPool` pool size is not configurable** — hardcoded to max(8, cpus*2); should read from config.
10. **`SearchBaseActor` still calls `e.printStackTrace()`** — pre-existing but touched in this PR; should use the project's TelemetryManager/logger.
11. **`EnrichSpec` tests exercise only routing** — TestEnrichActor mocks away all logic; there are no tests that verify the Kafka event structure or the ES lookup. See inline comment.

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingCache.java
```diff
@@ -0,0 +1,82 @@
...
+        synchronized (store) {
+            Entry e = store.get(key);
+            if (e == null) return null;
+            if (System.currentTimeMillis() - e.timestampMs > ttlMillis) {
+                store.remove(key);
+                return null;
+            }
+            return e.vector;
+        }
+    }
```
sntiwari1, line 46: **Bug: returns mutable reference — cache can be corrupted.**

`SemanticQueryStrategy.build()` receives this array and could mutate it (e.g. if quantization ever operates in-place, or if any caller modifies the returned float[] for post-processing). The next request for the same key then gets the corrupted vector with no error.

Fix: return a defensive copy:
```java
return e.vector.clone();
```

The allocation cost is negligible vs. the embedding HTTP call that was avoided.

search-api/search-service/conf/application.conf
```diff
@@ -319,10 +319,58 @@ ekstepPlatformApiUserId="search-service"
...
 
-compositesearch.index.name="kp_audit_log"
+compositesearch.index.name="compositesearch"
 
 audit.index.name="kp_audit_log"
 
+# Semantic search configuration. Disabled by default; turn on per env.
+# Vectors at index time are produced by content-embedding-job — the values
+# under embedding/quantization here must match the job's config.
+semantic_search {
```
sntiwari1, line 329: **High: `enabled = true` ships as the committed default.**

The PR description says this should be a feature flag defaulting to `false` so operators can enable it per environment. With `true` here, any deployment that pulls this config as-is enables semantic search, which will call the embedding service and fail at startup if `api_key` is empty or the E5 host is unreachable.

Change to:
```hocon
enabled = false
enabled = ${?SEMANTIC_SEARCH_ENABLED}
```

search-api/search-service/conf/application.conf
```diff
@@ -319,10 +319,58 @@ ekstepPlatformApiUserId="search-service"
 content.tagging.property=["subject","medium"]
 search.payload.log_enable=true
 
-compositesearch.index.name="kp_audit_log"
+compositesearch.index.name="compositesearch"
```
sntiwari1, line 322: **Breaking default config change — needs explicit callout.**

`compositesearch.index.name` was `kp_audit_log` and is now `compositesearch`. If this is a bug fix (the old value looked wrong — audit log ≠ composite search index), that's valid. But:

1. This should be mentioned explicitly in the PR description so operators know they may need to update their env-level config overrides.
2. Any existing deployment that relied on the `kp_audit_log` default (even mistakenly) will silently query the wrong index after this change.

Please confirm this is intentional and note it in the PR body.

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
```diff
@@ -0,0 +1,167 @@
...
+        List<String> identifiers = parseIdentifiers(request);
+        String topic = Platform.config.hasPath("kafka.publish.request.topic")
+                ? Platform.config.getString("kafka.publish.request.topic")
+                : "sunbirddev.publish.job.request";
+
+        org.sunbird.search.client.ElasticSearchUtil.initialiseESClient(
+                SearchConstants.COMPOSITE_SEARCH_INDEX,
+                Platform.config.getString("search.es_conn_info"));
+
+        return scala.concurrent.Future.apply(new scala.runtime.AbstractFunction0<Response>() {
```
sntiwari1, line 75: **Medium: ES client initialized on every request.**

`ElasticSearchUtil.initialiseESClient()` creates (or re-creates) the underlying REST client on each call. This is expensive and may reset connection pools mid-flight if called concurrently.

Move this to a `@PostConstruct`-equivalent or a lazy `static` initializer (the pattern the rest of the codebase uses in actors). The actor's `preStart()` hook is the right place in Pekko:
```java
@Override
public void preStart() throws Exception {
    super.preStart();
    ElasticSearchUtil.initialiseESClient(
        SearchConstants.COMPOSITE_SEARCH_INDEX,
        Platform.config.getString("search.es_conn_info"));
}
```

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
```diff
@@ -0,0 +1,167 @@
...
+            @Override
+            public Response apply() {
+                try {
+                    List<String> rawDocs = org.sunbird.search.client.ElasticSearchUtil
+                            .getMultiDocumentAsStringByIdList(SearchConstants.COMPOSITE_SEARCH_INDEX, identifiers);
+
```
sntiwari1, line 81: **Medium: blocking I/O on `ExecutionContext.Implicits$.MODULE$.global()` — can starve the dispatcher.**

The lambda contains two blocking calls: `getMultiDocumentAsStringByIdList` (ES HTTP) and the Kafka `send` loop. The global Scala EC is a `ForkJoinPool` not designed for blocking I/O; under load, this exhausts pool threads and stalls other futures.

This PR already introduces `SemanticEmbeddingPool` for the same reason. Either reuse it or create a dedicated IO pool for EnrichActor:
```java
return Future.apply(() -> { /* ... */ }, SemanticEmbeddingPool.context());
```
Or wire Pekko's `blocking-io-dispatcher` from `application.conf`.

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
```diff
@@ -0,0 +1,167 @@
...
+                    response.put("count", succeeded.size());
+                    response.put("identifiers", succeeded);
+                    response.put("failed", failed);
+                    return response;
+                } catch (ClientException e) {
+                    throw e;
+                } catch (Exception e) {
+                    throw new RuntimeException("EnrichActor: failed to fetch documents from index", e);
+                }
+            }
```
sntiwari1, line 128: **Medium: no cap on identifier count — unbounded ES multi-get + Kafka loop.**

A caller can POST thousands of identifiers, resulting in:
- One large ES multi-get request that can time out or put pressure on the cluster.
- A serial Kafka send loop of the same size, holding the thread for seconds.

Add a configurable maximum:
```java
int maxIds = Platform.config.hasPath("enrich.max_identifiers")
    ? Platform.config.getInt("enrich.max_identifiers") : 100;
if (identifiers.size() > maxIds)
    throw new ClientException("ERR_TOO_MANY_IDENTIFIERS",
        "Maximum " + maxIds + " identifiers per request, got: " + identifiers.size());
```

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
```diff
@@ -0,0 +1,240 @@
...
+            if (savedProps != null) {
+                for (Map p : savedProps) {
+                    Object pn = p.get("propertyName");
+                    if (!"*".equals(pn)) filterProps.add(p);
+                }
+            }
+            dto.setProperties(filterProps);
+            QueryBuilder existing = processor.buildTextQuery(dto);
+            if (existing instanceof BoolQueryBuilder) {
+                BoolQueryBuilder bx = (BoolQueryBuilder) existing;
```
sntiwari1, line 175: **Assumption about `should` semantics needs a guard or comment.**

Promoting `should` clauses to `filter` is safe here only because the comment above correctly describes where they come from (implicit-filter visibility wraps with `minimum_should_match=1`). But this assumption is fragile: if any caller ever adds a soft-constraint `should` and forgets to clear `softConstraints` before calling `buildTextQuery`, it silently becomes a hard filter exclusion in semantic mode.

The `softConstraints` clearing above is the defense, but it's separated by 50+ lines from this promotion. Suggest adding a defensive assertion, or an explicit comment tying the two together:
```java
// INVARIANT: softConstraints are cleared above before buildTextQuery(),
// so any should() here is exclusively from implicit-filter visibility wraps
// (getSearchQuery's origFilterQry / implFilterQuery). These ARE visibility
// constraints and must match, so promoting to filter() is correct.
// If this assertion ever fails, a scoring should() leaked through:
assert bx.minimumShouldMatch() == null || bx.minimumShouldMatch().equals("1");
for (QueryBuilder s : bx.should()) finalBool.filter(s);
```

search-api/search-core/src/main/java/org/sunbird/search/processor/SemanticEmbeddingPool.java
```diff
@@ -0,0 +1,30 @@
...
+    private static volatile ForkJoinPool pool;
+    private static volatile ExecutionContext context;
+
+    private SemanticEmbeddingPool() { }
+
+    static ExecutionContext context() {
+        if (context == null) {
+            synchronized (SemanticEmbeddingPool.class) {
+                if (context == null) {
+                    int parallelism = Math.max(8, Runtime.getRuntime().availableProcessors() * 2);
```
sntiwari1: **Low: pool size should be configurable.**

`max(8, cpus * 2)` is a reasonable default but in a heavily containerized environment (1-2 CPUs allocated) this always resolves to 8 threads, and in a high-CPU host it could grow to 64+ for a pool that only makes HTTP calls (where 8–16 is usually optimal). 

Suggest reading from config with a sensible fallback:
```java
int parallelism = Platform.config.hasPath("semantic_search.embedding_pool_size")
    ? Platform.config.getInt("semantic_search.embedding_pool_size")
    : Math.max(8, Runtime.getRuntime().availableProcessors() * 2);
```

search-api/search-service/test/controllers/EnrichSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+import play.api.test.Helpers._
+
+@RunWith(classOf[JUnitRunner])
+class EnrichSpec extends BaseSpec {
+
+    "Enrich API" should {
+
+        "return success for valid enrich request" in {
+            val response = post("/v3/enrich", """{"request": {"identifiers": ["do_123"]}}""")
+            isOK(response)
```
sntiwari1, line 14: **Tests validate HTTP routing only — actor logic is fully mocked.**

`TestEnrichActor` in `TestModule` replaces the real actor, so these tests confirm that:
- The route is wired (`/v3/enrich` → `triggerEnrich()`)
- Request validation rejects empty/missing `identifiers`

They do NOT cover:
- The ES lookup for identifier existence
- The Kafka event structure (correct `eid`, `edata.action`, `objectType` derivation)
- Partial-failure handling (some identifiers succeed, some fail Kafka send)
- The `failed` field in the response

At minimum, add a unit test for `EnrichActor` directly (using Pekko TestKit with mocked KafkaClient and ElasticSearchUtil) that verifies the Kafka event payload shape.

search-api/search-actors/src/main/java/org/sunbird/actors/SearchBaseActor.java
```diff
@@ -347,6 +347,15 @@ public SearchDTO getSearchDTO(Request request) throws Exception {
             if (fuzzySearch != null) {
                 searchObj.setFuzzySearch(fuzzySearch);
             }
+
+            // Parse search_mode (default text) and the optional semantic block.
+            // Validation of values happens in the strategy layer.
+            String searchMode = (String) req.get(SearchConstants.search_mode);
+            searchObj.setSearchMode(searchMode);
+            Object semanticObj = req.get(SearchConstants.semantic);
```
sntiwari1, line 355: **Nit: `e.printStackTrace()` should use structured logging.**

This is pre-existing but was touched in this PR. Use `TelemetryManager.error()` (used elsewhere in this file) so the exception lands in the platform's log aggregation instead of stderr:
```java
TelemetryManager.error("SearchBaseActor: invalid param cast: " + e.getMessage(), e);
```

search-api/search-core/src/main/java/org/sunbird/search/strategy/QueryStrategyFactory.java
```diff
@@ -0,0 +1,79 @@
...
+                        "search_mode=" + key + " is not enabled on this server");
+            }
+            throw new ClientException(
+                    SearchConstants.ERR_INVALID_SEARCH_MODE,
+                    "Unsupported search_mode: " + mode);
+        }
+        return strategy;
+    }
+
+    // TODO: Add EmbeddingCircuitBreaker wrapping SemanticQueryStrategy.
```
sntiwari1, line 68: Good TODO comment — worth tracking this as a follow-up issue before this goes to production. Without the circuit breaker, a sustained embedding-service outage makes every semantic/hybrid request throw a 500 (RuntimeException from the HTTP timeout) until operators manually flip the feature flag. The breaker described here would degrade gracefully to text mode, which is the much better operator experience.

Consider opening a follow-up ticket and referencing it here: `// TODO(#ISSUE): Add EmbeddingCircuitBreaker...`

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingClientFactory.java
```diff
@@ -0,0 +1,80 @@
...
+    private static EmbeddingClient build() {
+        String service = Platform.config.hasPath("semantic_search.embedding_service")
+                ? Platform.config.getString("semantic_search.embedding_service")
+                : "openai";
+
+        if ("openai".equalsIgnoreCase(service)) {
+            EmbeddingClientConfig cfg = EmbeddingClientConfig.builder()
+                    .serviceName("openai")
+                    .dimensions(getInt("semantic_search.openai.dimensions", 1536))
+                    .timeoutSeconds(getInt("semantic_search.openai.timeout", 5))
```
sntiwari1, line 50: Good that `OpenAIEmbeddingClient` constructor validates the key is non-empty (throws `IllegalArgumentException`). The empty string default here means the error is deferred to the first semantic request, not at startup. Consider adding an eager validation here when `semanticEnabled=true` so misconfiguration is caught at boot rather than at query time:
```java
String key = getString("semantic_search.openai.api_key", "");
if (key.isEmpty())
    throw new ClientException("ERR_MISSING_API_KEY",
        "semantic_search.openai.api_key or OPENAI_API_KEY env var must be set");
```
(Non-blocking suggestion — current behaviour is fine if you prefer lazy validation.)

COMMIT_RANGE_ANALYSIS.md
chethann007: I have removed the file, thanks for pointing it out!

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingCache.java
chethann007: Applied the changes that were suggested, thanks!

search-api/search-service/conf/application.conf
chethann007: Intentional — fail-fast on misconfiguration is better than silently disabled. Operators override per environment if they want it off.

search-api/search-service/conf/application.conf
chethann007: Updated the PR Description regarding this change. Thanks for the suggestion.

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
chethann007: Applied the suggested changes, thanks for the suggestion!

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
chethann007: Applied the changes, thanks for the suggestion!

search-api/search-actors/src/main/java/org/sunbird/actors/EnrichActor.java
chethann007: Applied the changes, thanks for the suggestion!

search-api/search-core/src/main/java/org/sunbird/search/strategy/SemanticQueryStrategy.java
chethann007: Applied the changes, thanks for the suggestion!

search-api/search-core/src/main/java/org/sunbird/search/processor/SemanticEmbeddingPool.java
chethann007: Applied the suggested changes, thanks!

search-api/search-service/test/controllers/EnrichSpec.scala
chethann007: Added new unit tests, thanks for the suggestion!

search-api/search-core/src/main/java/org/sunbird/search/embedding/EmbeddingClientFactory.java
chethann007: - Current behavior: OpenAIEmbeddingClient constructor validates on first use and throws `llegalArgumentException` with a clear message. That's fine.

- Eager validation at `EmbeddingClientFactory.build()` would make every deployment fail at startup if the key isn't set — including environments that have semantic search enabled in config but haven't configured the embedding service yet, or environments where the key comes from a secret injected at runtime after app start.

- Lazy validation means the error surfaces on the first semantic request with a clear exception — not silent, not harmful, easier to operate.

search-api/search-actors/src/main/java/org/sunbird/actors/SearchBaseActor.java
chethann007: Pre-existing code, out of scope for this PR. Will track separately.

search-api/search-core/src/main/java/org/sunbird/search/strategy/QueryStrategyFactory.java
chethann007: Will track as a follow-up issue.

pallakartheekreddy (approved):

PR #1240, opened 2026-05-28
feat: Implement SCORM package ingestion, playback, and integrity validation

romitshah02:
Summary of Change
Implemented end-to-end SCORM package ingestion and playback support for both single and multi sco. This includes backend ingestion logic (ScormMimeTypeMgrImpl), editor support for MIME detection and binary upload, and a SCORM API shim in the player. Included an integrity check to verify the existence of the launchFile specified in imsmanifest.xml.

Motivation
To enable compatibility with standard SCORM learning modules within the platform, ensuring robust package ingestion and reliable playback.

Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

How Has This Been Tested?
- [x] Verified E2E flow (Authoring -> Ingestion -> Playback) locally.
- [x] Verified invalid SCORM packages now throw ClientException during ingestion.
- [x] Verified valid SCORM packages ingest and playback correctly.

Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * SCORM support: package upload, extraction, artifact storage, and automatic launch-file detection; review and publish flows.

* **Bug Fixes / Validation**
  * Rejects invalid SCORM packages and missing/invalid launch files with clear errors.

* **Security**
  * Extraction hardened to prevent path traversal/Zip-Slip and ensure safe launch-file resolution.

* **Tests**
  * Added unit tests covering SCORM upload validation and failure cases.

* **Schema**
  * Content schema now allows the SCORM MIME type.

* **Chores / Dev**
  * Local seed-data tooling and README updates.

* **Infrastructure**
  * Compose/VM and CI updates: OpenSearch migration and simplified build workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Sunbird-Knowlg/knowledge-platform/pull/1234?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,135 @@
...
+					getValidatedLaunchFile(extractionBasePath, scoList.head("href"))
+				}
+
+                node.getMetadata.put("scoList", mapper.writeValueAsString(scoList))
+
+				val urls: Array[String] = blocking {
+					uploadArtifactToCloud(uploadFile, objectId, filePath)
+				}
+				node.getMetadata.put("s3Key", urls(IDX_S3_KEY))
+				node.getMetadata.put("artifactUrl", urls(IDX_S3_URL))
```
sntiwari1: **Bug: `blocking` wrappers do not actually make this Future-based code non-blocking.**

`blocking { extractPackage(...) }` is only meaningful inside a `Future { ... }` on the global ExecutionContext — it hints to the thread pool to spawn an extra thread. But here `extractPackage`, `uploadArtifactToCloud`, and `extractPackageInCloud` are all called synchronously (outside any `Future`), so they block the calling thread regardless of the `blocking` hint.

The final result is wrapped in `Future { Map(...) }` — a no-op that just wraps an already-computed value.

Compare to `H5PMimeTypeMgrImpl`, which also calls these methods synchronously without `blocking`. Either follow the same pattern consistently (plain synchronous calls, no `blocking`) or, if true async is desired, wrap the entire body in `Future { blocking { ... } }`.

```scala
// Current (misleading): blocking { extractPackage(...) } called on calling thread
// Fix — if async is needed:
Future {
  blocking {
    extractPackage(uploadFile, extractionBasePath)
    val scoList = getScoList(getSecureXml(manifestFile))
    ...
  }
}
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,135 @@
...
+            TelemetryManager.error("ERR_INVALID_FILE:: " + "Invalid SCORM package structure: imsmanifest.xml not found! with file name: " + uploadFile.getName)
+            throw new ClientException("ERR_INVALID_FILE", "Invalid SCORM package: imsmanifest.xml is missing!")
+        }
+    }
+
+    private def getValidatedLaunchFile(extractionBasePath: String, launchFile: String): String = {
+        if (launchFile.isEmpty) {
+            throw new ClientException("ERR_INVALID_FILE", "Invalid launch file path!")
+        }
+
```
sntiwari1: **Bug: temp directory is never cleaned up on the `else` branch (invalid package).**

The `finally { delete(new File(extractionBasePath)) }` block only runs for the `if` branch. When `isValidPackageStructure` returns `false`, `extractionBasePath` has already been created by `getBasePath(objectId)` but the `finally` is never reached. Leaked temp directories accumulate on disk under high load.

Move the cleanup to wrap the entire block:

```scala
val extractionBasePath = getBasePath(objectId)
try {
  if (isValidPackageStructure(...)) { ... }
  else { throw new ClientException(...) }
} finally {
  delete(new File(extractionBasePath))
}
```

sntiwari1 (commented):
## SCORM Ingestion PR Review

This review covers the core Scala implementation files. The SCORM feature is a valuable addition; below are correctness/security/design issues that should be addressed before merge.

sntiwari1 (commented):
BaseMimeTypeManager.scala — Dead code and misleading indentation in `extractPackage` (lines ~143-156)

The PR adds a second path-traversal check using `baseDirPath`:

```
val baseDirPath = Paths.get(basePath).toAbsolutePath.normalize()
for (entry <- ...) {
    val path = resolvedBaseDir.resolve(entry.getName).normalize()
    if (!path.startsWith(resolvedBaseDir))   // existing check
        throw new ClientException(...)
    if (!path.startsWith(baseDirPath)) {     // new PR check — dead code
        throw new ClientException(...)
    }
```

`resolvedBaseDir` is `baseDir.toRealPath()` (canonical, symlink-resolved). `baseDirPath` is `Paths.get(basePath).toAbsolutePath.normalize()` — these resolve to the same directory. Because any path that passes the first check (`startsWith(resolvedBaseDir)`) will also satisfy the second (`startsWith(baseDirPath)`), the second block is unreachable dead code.

Additionally the indentation in the PR diff is broken — the `for` loop is at an incorrect nesting level, making the structure misleading.

Recommendation: Remove the `baseDirPath` variable and its check entirely. The existing `resolvedBaseDir` check is sufficient and correct.

sntiwari1 (commented):
ScormMimeTypeMgrImpl.scala — Mutable `val mapper` on a class instance instead of a companion object

```scala
class ScormMimeTypeMgrImpl(implicit ss: StorageService) extends ... {
    val mapper = new ObjectMapper()
    mapper.registerModule(DefaultScalaModule)
```

`ObjectMapper` is thread-safe after configuration but the `registerModule` call happens in the constructor body. A new `ScormMimeTypeMgrImpl` is instantiated once and shared (see `MimeTypeManagerFactory`), so while this is not a concurrency bug in practice, it is an anti-pattern. More importantly `ObjectMapper` is an expensive object to construct.

Recommendation: Move `mapper` to a companion object so it is created exactly once for the JVM lifetime:

```scala
object ScormMimeTypeMgrImpl {
  private val mapper: ObjectMapper = new ObjectMapper()
    .registerModule(DefaultScalaModule)
}
```

This also removes the need for a separate `registerModule` call on a `val`.

sntiwari1 (commented):
ScormMimeTypeMgrImpl.scala — `getScoList` uses an unsafe `.asInstanceOf` cast that hides type errors at compile time

```scala
}).toList.asInstanceOf[List[Map[String, String]]]
```

The `.map(...)` already returns `List[Map[String, String]]` at the call site — Scala infers this from the literal `Map("identifier" -> ..., "title" -> ..., "href" -> ...)` expressions whose values are all `String`. The explicit `.asInstanceOf` cast suppresses that inference, making compiler type-checking blind to future regressions where a value might not be `String`.

Recommendation: Remove the cast entirely and let the type be inferred:

```scala
private def getScoList(xml: Elem): List[Map[String, String]] = {
    (xml \\ "item").filter(item => (item \@ "identifierref").nonEmpty).map { item =>
        val ref   = item \@ "identifierref"
        val title = (item \ "title").text
        val href  = (xml \\ "resource").find(res => (res \@ "identifier") == ref)
                                        .map(_ \@ "href").getOrElse("")
        Map("identifier" -> (item \@ "identifier"), "title" -> title, "href" -> href)
    }.toList
}
```

sntiwari1 (commented):
ScormMimeTypeMgrImpl.scala — `upload` throws a synchronous exception in the `else` branch, breaking the `Future`-returning contract

```scala
override def upload(...): Future[Map[String, AnyRef]] = {
    validateUploadRequest(...)
    if (isValidPackageStructure(...)) {
        ...
    } else {
        TelemetryManager.error(...)
        throw new ClientException(...)  // synchronous throw inside a Future-returning method
    }
}
```

When a method is declared to return `Future[T]`, callers expect all failures to be communicated through a failed `Future`, not through a synchronous exception. If a caller wraps this in a `Future { scormMgr.upload(...) }.flatten`, the synchronous throw is caught by that outer `Future`. But if any caller calls `upload(...)` directly and chains `.map`/`.flatMap` on the result, the synchronous exception bypasses all `recover`/`recoverWith` handlers on the future chain — the same issue as `validateUploadRequest`.

Recommendation: Wrap the failure in a failed `Future`:

```scala
} else {
    TelemetryManager.error(...)
    Future.failed(new ClientException("ERR_INVALID_FILE", "Invalid SCORM package: imsmanifest.xml is missing!"))
}
```

Note: `validateUploadRequest` also throws synchronously, so wrapping that call too would make this fully consistent.

sntiwari1 (commented):
ScormMimeTypeMgrImpl.scala — `getSecureXml` overrides `parser` in an anonymous class but the field is `def parser`, not a stable `val`

```scala
val xmlLoader = new XMLLoader[Elem] {
    override def parser = saxParser
}
xmlLoader.loadFile(manifestFile)
```

`XMLLoader.parser` is a `def` that creates a new `SAXParser` on every call in the default implementation. The override here correctly returns the pre-configured `saxParser`, but because `parser` is a `def` (not a `val`), there is a subtle contract issue: if `loadFile` calls `parser` more than once internally (e.g., for error-recovery paths), a different `SAXParser` instance would normally be returned each time — but here the same instance is reused. `SAXParser` is not thread-safe and is not designed to be called twice without reset.

More importantly, the Scala standard library `XML.loadFile` method is not hardened against XXE and should not be used without this pattern — which the author correctly implements. However, the approach is fragile because it relies on internal behaviour of `XMLLoader`.

Recommendation: Use `saxParser.getXMLReader` and configure it directly, then parse via `scala.xml.XML.withSAXParser(saxParser).loadFile(manifestFile)` which is the idiomatic and stable API for supplying a custom SAX parser to the Scala XML library.

sntiwari1 (commented):
schemas/content/1.0/schema.json — Inconsistent indentation for the new SCORM MIME type entry

```json
        "application/vnd.ekstep.plugin-archive",
+          "application/vnd.ekstep.scorm-archive",
        "application/vnd.ekstep.h5p-archive",
```

The added line uses extra leading spaces compared to its neighbors. This is a minor cosmetic issue but can cause unnecessary noise in future diffs and may indicate an accidental copy-paste. Please align indentation to match the surrounding array entries (two extra spaces are being inserted).

sntiwari1 (commented):
ScormMimeTypeMgrImplTest.scala — Test for multi-SCO package parses `scoList` with raw `classOf[List[Map[String, String]]]` which silently produces wrong type at runtime

```scala
val scoList = mapper.readValue(scoListJson, classOf[List[Map[String, String]]])
```

Due to JVM type erasure, `classOf[List[Map[String, String]]]` is the same as `classOf[List[_]]`. Jackson deserializes it as `List[java.util.LinkedHashMap[String, Object]]`, not `List[Map[String, String]]`. The test assertions `sco("identifier") == "item1"` happen to work only because `LinkedHashMap` is also a `java.util.Map` and Scala can call `.apply` on it, but the Scala type annotation `List[Map[String, String]]` is a lie — at runtime this is a `List[java.util.LinkedHashMap]`.

Use a `TypeReference` to deserialize correctly:

```scala
import com.fasterxml.jackson.core.`type`.TypeReference
val scoList = mapper.readValue(scoListJson,
    new TypeReference[java.util.List[java.util.Map[String, String]]]() {})
scoList.size shouldBe 2
```

Or alternatively use `mapper.readTree(scoListJson)` for assertion purposes in tests.

sntiwari1 (commented):
## Code Review Summary — PR #1240: SCORM ingestion, playback & integrity validation

**Overall:** Good foundational implementation with solid security thinking (XXE protection, Zip-Slip prevention, path-traversal guard). A few correctness issues need attention before merge.

---

### Bugs / Correctness (must fix)

**1. `launchFile` not persisted to the graph node**
`scoList` is written to `node.getMetadata` AND returned in the result map (double-persisted), but `launchFile` is only in the result map. Be consistent: either add `node.getMetadata.put("launchFile", launchFile)` alongside the `scoList` mutation, or remove both direct mutations and let the result map drive persistence for all fields (like `s3Key`/`artifactUrl` do).

**2. `scoList` JSON-serialised twice**
`ScormMimeTypeMgrImpl.mapper.writeValueAsString(scoList)` is called on two separate lines. Assign to a `val` first to avoid the redundant work.

**3. Indentation regression in `BaseMimeTypeManager.extractZipFile`**
The `for` loop body was un-indented one level, making it visually appear outside the `try` block. The logic is correct but the formatting is misleading — restore the original indentation.

---

### Design / Gaps (should fix)

**4. Only the first SCO href is validated**
For multi-SCO packages, only `scoList.head`'s launch file is checked against the zip contents. Subsequent SCO hrefs go unverified and will cause runtime failures during playback. Either validate all hrefs up-front, or document the limitation clearly.

**5. No zip-bomb protection**
No limit on decompressed entry size or entry count. A crafted zip could exhaust disk or heap. Consider a threshold (e.g. max 500 entries, max 500 MB uncompressed) consistent with what H5P/HTML managers allow.

**6. Missing test: empty scoList (no items with `identifierref`)**
The path where `getScoList` returns an empty list and the `ClientException("ERR_INVALID_FILE", "No SCOs found...")` is thrown has no test coverage.

---

### Minor / Style

- The return `Map[String, AnyRef](...)` on one long line in `upload` should be split for readability.
- `review` and `publish` validation messages wrap text in `[...]` brackets — remove for consistency with other managers.
- `MimeTypeManagerFactory.scala` and `MimeTypeManagerFactoryTest.scala` are missing a trailing newline (causes noisy future diffs).

---

### What is done well
- XXE protection in `getSecureXml` is comprehensive (external-general-entities, external-parameter-entities, disallow-doctype-decl, load-external-dtd all disabled).
- Path-traversal guard in `getValidatedLaunchFile` via virtual-base normalize correctly rejects `../` sequences and absolute paths.
- `blocking {}` wrapper in both SCORM and HTML managers avoids thread-pool starvation on blocking I/O.
- Temp file cleanup on URL-based upload (finally + `FileUtils.deleteDirectory`) is correct.
- Test coverage across happy paths and three attack vectors is a solid start.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,135 @@
...
+                            manifestStream.close()
+                        }
+
+                        if (scoList.isEmpty) {
+                            throw new ClientException("ERR_INVALID_FILE", "No SCOs found in imsmanifest.xml!")
+                        }
+
+                        val launchFile = getValidatedLaunchFile(zipFile, scoList.head.getOrElse("href", ""))
+
+                        node.getMetadata.put("scoList", ScormMimeTypeMgrImpl.mapper.writeValueAsString(scoList))
```
sntiwari1: **Bug: `launchFile` not persisted + `scoList` double-serialised**

`scoList` is written to `node.getMetadata` here *and* returned in the result map on the last `Map[...]` line — it ends up persisted twice. But `launchFile` is only in the result map, making persistence inconsistent.

Suggested fix — store both via the result map and remove the direct mutation here, consistent with how `s3Key`/`artifactUrl` are handled:

```scala
val scoListJson = ScormMimeTypeMgrImpl.mapper.writeValueAsString(scoList)
// remove: node.getMetadata.put("scoList", scoListJson)
...
Map[String, AnyRef](
  "identifier"  -> objectId,
  "artifactUrl" -> urls(IDX_S3_URL),
  "s3Key"       -> urls(IDX_S3_KEY),
  "size"        -> getFileSize(uploadFile).asInstanceOf[AnyRef],
  "launchFile"  -> launchFile,
  "scoList"     -> scoListJson
)
```
This also removes the second `writeValueAsString(scoList)` call.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,135 @@
...
+                            getScoList(getSecureXml(manifestStream))
+                        } finally {
+                            manifestStream.close()
+                        }
+
+                        if (scoList.isEmpty) {
+                            throw new ClientException("ERR_INVALID_FILE", "No SCOs found in imsmanifest.xml!")
+                        }
+
+                        val launchFile = getValidatedLaunchFile(zipFile, scoList.head.getOrElse("href", ""))
```
sntiwari1: **Design gap: only the first SCO href is validated**

For multi-SCO packages this only checks `scoList.head`. The remaining SCOs are returned in `scoList` but their launch files are never verified to exist in the zip. A player sequencing through all SCOs will hit a runtime error for any missing subsequent href.

Either validate all hrefs up-front:
```scala
scoList.foreach { sco =>
  getValidatedLaunchFile(zipFile, sco.getOrElse("href", ""))
}
val launchFile = scoList.head.getOrElse("href", "")
```

Or document the intentional limitation in a comment so it is not silently broken.

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImpl.scala
```diff
@@ -0,0 +1,135 @@
...
+                        node.getMetadata.put("scoList", ScormMimeTypeMgrImpl.mapper.writeValueAsString(scoList))
+
+                        val urls: Array[String] = uploadArtifactToCloud(uploadFile, objectId, filePath)
+                        
+                        node.getMetadata.put("s3Key", urls(IDX_S3_KEY))
+                        node.getMetadata.put("artifactUrl", urls(IDX_S3_URL))
+                        
+                        extractPackageInCloud(objectId, uploadFile, node, "snapshot", false)
+                        
+                        Map[String, AnyRef]("identifier" -> objectId, "artifactUrl" -> urls(IDX_S3_URL), "size" -> getFileSize(uploadFile).asInstanceOf[AnyRef], "s3Key" -> urls(IDX_S3_KEY), "launchFile" -> launchFile, "scoList" -> ScormMimeTypeMgrImpl.mapper.writeValueAsString(scoList))
```
sntiwari1: **Style: single-line result map is hard to scan**

This is one of the longest lines in the module. Please break it across multiple lines:
```scala
Map[String, AnyRef](
  "identifier"  -> objectId,
  "artifactUrl" -> urls(IDX_S3_URL),
  "s3Key"       -> urls(IDX_S3_KEY),
  "size"        -> getFileSize(uploadFile).asInstanceOf[AnyRef],
  "launchFile"  -> launchFile,
  "scoList"     -> scoListJson
)
```

platform-modules/mimetype-manager/src/main/scala/org/sunbird/mimetype/mgr/BaseMimeTypeManager.scala
```diff
@@ -140,17 +140,18 @@ class BaseMimeTypeManager(implicit ss: StorageService) {
 		val resolvedBaseDir = baseDir.toRealPath()
 		val zipFile = new ZipFile(file)
 		try {
```
sntiwari1, line 142: **Indentation regression: `for` loop body looks like it's outside the `try` block**

The surrounding `try {` / `} finally {` has `for` at the same indentation level as `}`, making it visually ambiguous whether the loop is inside the `try`. The logic is correct (Scala braces determine scope, not indentation), but this misleads readers. Restore the original indentation so `for` is nested one level inside `try {`:

```scala
try {
  for (entry <- zipFile.entries().asScala) {
    val path = resolvedBaseDir.resolve(entry.getName).normalize()
    if (!path.startsWith(resolvedBaseDir))
      throw new ClientException("ERR_INVALID_ZIP_ENTRY",
        "Zip entry attempts path traversal: " + entry.getName)
    if (entry.isDirectory) Files.createDirectories(path)
    else {
      Files.createDirectories(path.getParent)
      Files.copy(zipFile.getInputStream(entry), path)
    }
  }
} finally {
  zipFile.close()
}
```

platform-modules/mimetype-manager/src/test/scala/org/sunbird/mimetype/mgr/impl/ScormMimeTypeMgrImplTest.scala
```diff
@@ -0,0 +1,149 @@
...
+            case scala.util.Failure(e: ClientException) =>
+                e.getErrCode shouldBe "VALIDATOR_ERROR"
+                scala.util.Success(succeed)
+            case scala.util.Failure(e) =>
+                scala.util.Failure(new Exception(s"Expected ClientException, got ${e.getClass.getName}: ${e.getMessage}", e))
+            case scala.util.Success(_) =>
+                scala.util.Failure(new Exception("Expected ClientException, but review succeeded"))
+        }
+    }
+}
```
sntiwari1, line 149: **Missing test: no SCOs in manifest (empty scoList)**

The path where `imsmanifest.xml` is present but contains no `<item>` elements with an `identifierref` attribute — causing `scoList.isEmpty` → `ClientException("ERR_INVALID_FILE", "No SCOs found...")` — has no coverage. Please add:

```scala
"upload" should "throw ClientException when manifest contains no SCOs" in {
  val manifest = "<manifest><organizations/><resources/></manifest>"
  val file = createZip(Map("imsmanifest.xml" -> manifest))
  Future { scormMgr.upload("do_1", getNode(), file, None, UploadParams()) }.flatten.transform {
    case scala.util.Failure(e: ClientException) =>
      e.getErrCode shouldBe "ERR_INVALID_FILE"
      e.getMessage should include("No SCOs")
      FileUtils.deleteQuietly(file)
      scala.util.Success(succeed)
    case _ =>
      FileUtils.deleteQuietly(file)
      scala.util.Failure(new Exception("Expected ClientException for empty SCO list"))
  }
}
```

pallakartheekreddy (approved):

PR #1243, opened 2026-06-04
feat: Enable FTB, MTF, Sequence, and Reorder question types in QuML 1.1 with i18n support

sntiwari1:
## Summary

Implements support for four new QuML 1.1 question types — Fill-in-the-Blank (FTB),
Match The Following (MTF), Sequence (SEQ), and Reorder (REO) — along with a
`?lang=<code>` query parameter for i18n-aware question reads.

### Changes

**Schema (3 files — additive only, backward compatible)**
- `schemas/question/1.1/schema.json` + `test_schema/question/1.1/schema.json`:
  added `"match"`, `"order"` to `interactionTypes`; `"SEQ"`, `"REO"` to `qType`;
  `"ordered"` to `responseDeclaration.cardinality`; `"map"` to `responseDeclaration.type`;
  `"responseProcessing"`, `"offline"`, `"external"` to `scoringMode`; typed `interactions`
  schema with `oneOf` flat-array/left-right and `interactionOption` definitions block;
  relaxed `responseDeclaration.additionalProperties` to `true` (unblocks `maxScore`).
- `schemas/questionset/1.1/schema.json`: extended `scoringMode` enum.

**Definition scripts (5 files)**
- Updated `FTB_QUESTION.sh` with typed schema constraints.
- Created `Match_The_Following_Question.sh`, `Sequence_Question.sh`,
  `Reorder_Question.sh`.
- Appended MTF / SEQ / REO master category entries to `master_category_create`.

**AssessmentV5Manager.scala**
- Added `import scala.util.Try`.
- `getAnswer`: added defensive `case _ => ""` to prevent `MatchError` on new
  cardinality values.
- `processInteractions` + `processResponseDeclaration`: guarded v1.0-only
  migration transforms behind `version < 1.1` checks (defence-in-depth; both
  methods were already guarded at the call-site).
- `validateQuestionNodeForReview`: extended from presence-only to type-specific
  structural validation; dispatches to `validateFTB`, `validateMTF`, or
  `validateOrdered` based on `qType`.
- `getQuestionMetadata`: added optional `lang: String = ""` parameter and
  `filterByLanguage` call.
- New private helpers: `validateFTB`, `validateMTF`, `validateOrdered`,
  `filterByLanguage`, `filterOptionLabels`, `systemDefaultLang`,
  `i18nTopLevelFields`.

**QuestionController.v5 + QuestionActor.v5**
- Controller `read`/`privateRead`: extract `?lang=` from HTTP query string and
  forward to request map.
- Actor `read`: extract `lang` from request and pass to `getQuestionMetadata`.

**Tests**
- New `AssessmentV5ManagerTest.scala` — 18 unit tests covering all validate
  helpers, language filter, and v1.0 guard behaviour.
- `scripts/test-new-question-types.sh` — end-to-end test script (15 cases):
  create + review for each type, bilingual lang= filter, and validation-rejection
  scenarios.
- `ft_question_questionset_v2.json` — 4 new Postman collections (FTB, MTF, SEQ,
  REO) with 14 test cases and 5 new collection variables.

### Test plan
- [ ] `mvn scalatest:test -pl assessment-api/assessment-actors -DwildcardSuites=org.sunbird.v5.managers.AssessmentV5ManagerTest` → 18/18 pass
- [ ] `./scripts/test-new-question-types.sh` against local stack → 15/15 pass
- [ ] Apply definition scripts: `scripts/definition-scripts/Match_The_Following_Question.sh`, `Sequence_Question.sh`, `Reorder_Question.sh`, `FTB_QUESTION.sh` against target environment
- [ ] Smoke test `POST /question/v5/create` for each of FTB / MTF / SEQ / REO
- [ ] Smoke test `GET /question/v5/read/:id?lang=hi` returns Hindi body for bilingual question
- [ ] Confirm existing MCQ / SA question types unaffected

### Notes
- All schema changes are additive; no existing valid payloads are rejected.
- v1.0 question read-path (`schemaVersion=1.0`) is fully isolated — new validation
  and lang-filter code paths are unreachable from it.
- Redis not required for basic operation (health check skips it gracefully).
- DB setup: run `docker/init-yugabyte.sh` for fresh environments to provision
  all required keyspaces and tables.

## Release Notes

* **New Features**
  * Added language-specific question retrieval capability.
  * Introduced support for new question types: Fill-in-the-Blanks, Match-The-Following, Sequence, and Reorder.
  * Enhanced validation for question metadata and interaction structures.

* **Tests**
  * Added test coverage for new question types and language-based filtering.


assessment-api/assessment-actors/src/main/scala/org/sunbird/v5/managers/AssessmentV5Manager.scala
```diff
@@ -72,17 +73,75 @@ object AssessmentV5Manager {
...
     else updatedMeta.keySet().removeAll(extFields)
     updatedMeta.put("identifier", node.getIdentifier.replace(".img", ""))
+    if (StringUtils.isNotBlank(lang)) filterByLanguage(updatedMeta, lang)
     updatedMeta
   }
 
+  private val systemDefaultLang: String = Platform.getString("platform.default.language", "en")
+  private val i18nTopLevelFields: Set[String] = Set("body", "instructions", "answer", "hints", "feedback", "solutions")
+
+  private[managers] def filterByLanguage(metadata: util.Map[String, AnyRef], lang: String): Unit = {
```
sntiwari1, line 121: **i18n filter may be a no-op for top-level fields in the real read path.**

`filterByLanguage` only rewrites a field when its value is a `util.Map` (`case m: util.Map[_, _]`). But `getQuestionMetadata` builds `metadata` from `NodeUtil.serialize` + `convertOneOfProps`, and `convertOneOfProps` only deserializes the *oneOf* props for the category. `body` (and `instructions`/`hints`/etc.) are not generally oneOf props — if they are persisted as a JSON **string** like `{"en":"...","hi":"..."}`, this match falls through to `case _ =>` and the `?lang=` filter silently does nothing.

The 18 unit tests all construct an in-memory `util.Map` for `body`, so they never exercise the serialized-string representation that the read path actually produces. Please verify against a real `GET /question/v5/read/:id?lang=hi` on a bilingual question (the PR test plan item) that `body` is in fact a `Map` at this point, and add a test that feeds the serialized form. If `body` arrives as a String, you'll need to `JsonUtils.deserialize` it (guarded by a try/catch) before resolving the language.

assessment-api/assessment-actors/src/main/scala/org/sunbird/v5/managers/AssessmentV5Manager.scala
```diff
@@ -133,24 +192,111 @@ object AssessmentV5Manager {
...
+      val right = opts.getOrDefault("right", new util.ArrayList[util.Map[String, AnyRef]]()).asInstanceOf[util.List[util.Map[String, AnyRef]]]
+      if (left.isEmpty)  errs += s"interactions.$key.options.left must be non-empty for MTF"
+      if (right.isEmpty) errs += s"interactions.$key.options.right must be non-empty for MTF"
+      if (correctMap != null && !left.isEmpty && !right.isEmpty) {
+        val leftVals  = left.asScala.map(_.getOrDefault("value", "").asInstanceOf[String]).toSet
+        val rightVals = right.asScala.map(_.getOrDefault("value", "").asInstanceOf[String]).toSet
+        correctMap.keySet().asScala.foreach(k => if (!leftVals.contains(k))  errs += s"correctResponse key '$k' not in left options")
+        correctMap.values().asScala.foreach(v => if (!rightVals.contains(v.asInstanceOf[String])) errs += s"correctResponse value '$v' not in right options")
+        val rhsUsed = correctMap.values().asScala.toList
+        if (rhsUsed.size != rhsUsed.toSet.size) errs += "correctResponse.value violates one-to-one constraint (duplicate RHS)"
```
sntiwari1, line 301: **MTF validation hard-codes a one-to-one (bijective) match.** Rejecting duplicate RHS values forbids legitimate many-to-one Match-The-Following questions (e.g. several countries mapping to the same continent), which QuML's `map` cardinality permits. If one-to-one is an intentional product constraint for this release, fine — but please confirm it's intended and document it, since it's stricter than the schema and not mentioned in the PR description.

assessment-api/qs-hierarchy-manager/src/main/scala/org/sunbird/managers/UpdateHierarchyManager.scala
```diff
@@ -243,7 +243,6 @@ object UpdateHierarchyManager {
         idMap += (nodeId -> identifier)
         metadata.put(HierarchyConstants.IDENTIFIER, identifier)
         metadata.put(HierarchyConstants.CODE, nodeId)
-        metadata.put(HierarchyConstants.VERSION_KEY, System.currentTimeMillis + "")
         metadata.put(HierarchyConstants.CREATED_ON, DateUtils.formatCurrentDate)
```
sntiwari1, line 247: **Out-of-scope and unexplained:** this PR removes `metadata.put(HierarchyConstants.VERSION_KEY, System.currentTimeMillis + "")` for newly created hierarchy child nodes. `versionKey` is used for optimistic-concurrency checks on subsequent updates, so dropping it at creation can change update behaviour for QuestionSet children. This change isn't mentioned in the PR summary and is unrelated to the FTB/MTF/SEQ/REO + i18n feature (the branch also has a "Revert HierarchyManager Changes" commit, suggesting churn here). Please either drop this from the PR or explain why it's needed and how it was verified.

search-api/search-core/src/main/java/org/sunbird/search/util/SearchInputValidator.java
```diff
@@ -34,7 +34,8 @@ public class SearchInputValidator {
             "subject", "artifactUrl", "downloadUrl", "variants", "pkgVersion",
             "audience", "license", "language", "lastPublishedOn", "identifier",
             "createdOn", "lastUpdatedOn", "resourceType", "appIcon", "posterImage",
-            "visibility", "compatibilityLevel", "mediaType", "origin", "rating"
+            "visibility", "compatibilityLevel", "mediaType", "origin", "rating",
+            "originData"
```
sntiwari1, line 38: Adding `originData` to the allowed search fields is unrelated to the QuML question-types feature and isn't mentioned in the PR description. If it's a needed dependency, please call it out in the summary; otherwise consider splitting it into its own PR to keep this one focused.

sntiwari1 (commented):
## Code Review — PR #1243: FTB / MTF / Sequence / Reorder question types + i18n

Thorough, well-structured PR. Schema changes are genuinely additive, the new validation helpers are readable and unit-tested, and gating the v1.0 migration transforms behind `version < 1.1` is the right call. CI is green except **SonarCloud Code Analysis (FAILURE)** — please check that before merge.

Below are the points worth addressing, roughly in priority order. Inline comments cover the specifics.

### 🔴 Should address before merge

1. **i18n `?lang=` filter may be a silent no-op for top-level fields** (`AssessmentV5Manager.filterByLanguage`). The helper only rewrites a field when its value is a `util.Map`, but the read path may deliver `body`/`instructions`/etc. as JSON *strings*. The 18 unit tests only feed in-memory Maps, so they don't catch this. Needs an end-to-end check on a real bilingual read (see inline comment).

2. **Removed security regression tests.** The old `AssessmentV5ManagerTest` contained explicit XSS escaping tests for `getAnswer` (cookie-theft `XSS-VULN-02`, JWT-theft payloads) plus the `getValidatedNodeForUpdateComment` null/empty/whitespace tests. The new file replaces the whole suite and drops all of these. `getAnswer` still runs the escaping logic, so the coverage shouldn't just be deleted — please re-add the XSS and comment-validation tests alongside the new ones.

3. **Two out-of-scope changes** not mentioned in the PR summary:
   - `UpdateHierarchyManager` drops `versionKey` assignment on new child nodes (optimistic-concurrency implications).
   - `SearchInputValidator` adds `originData` to allowed search fields.
   Either justify + verify them here or split into separate PRs.

### 🟡 Worth a look

4. **MTF one-to-one constraint** is stricter than the schema and forbids legitimate many-to-one matches — confirm it's intended (inline).

5. **`responseDeclaration.additionalProperties: true`** (schema). Relaxing this to unblock `maxScore` is a broad loosening — any typo'd property now passes validation. Consider explicitly whitelisting the few extra properties instead of opening it fully.

6. **`listQuestions` refactor** (wrapping `validateListRequest` in `Future(...).flatMap`) is a reasonable robustness improvement, but it's unrelated to this feature and unmentioned. Minor — just note it.

### 🟢 Nits / positives
- Good defensive `case _ => ""` in `getAnswer` to avoid `MatchError` on new cardinality values.
- Definition scripts and master-category entries are consistent across FTB/MTF/SEQ/REO.
- `validateFTB`/`validateMTF`/`validateOrdered` correctly anchor errors with the response key, which will make authoring errors easy to debug.

Overall: solid feature work. The blocking items are mostly about **verifying the i18n path actually fires end-to-end** and **not losing the existing security test coverage**; the scope-creep changes just need to be acknowledged or split out.

pallakartheekreddy (approved):

PR #1303, opened 2026-08-03
feat(content-api): Enrichment/Transcript object model, CRUD APIs, and Kafka wiring

chethann007:
## Summary

Adds a generic Enrichment/Transcript object model to knowlg-service and content-api — the platform-side counterpart to ai-pipeline's new `caption-generator`/`enrichment-router` Flink jobs. A Content node gets one Enrichment node holding typed AI-derived child objects (Transcript today), with generic (not per-object-type) CRUD APIs and Kafka events driving the transcription/multilingual-translation pipeline.

## Key changes

- New `Enrichment` and `Transcript` node schemas (`schemas/enrichment/1.0`, `schemas/transcript/1.0`) and the `Content -> Enrichment` graph relation.
- Generic Enrichment object CRUD (`create`/`update`/`approve`/`reject`) under `/content/v4/enrichment/object/*`, replacing an earlier per-object-type Transcript API design — same 4 endpoints work for any future child type.
- `GET /content/v4/enrichment/read/:id` and an opt-in `enrich` param on`content/v3|v4/read` that embeds a live Enrichment join.
- `BE_JOB_REQUEST` Kafka envelope emitted on Enrichment/Transcript lifecycle events, driving ai-pipeline's transcription and multilingual dispatch.
- `enriched.metadata` kafka topic support in `application.conf`.
- search-api: added `enrichment` to the exists-field allow-list so `{"exists": ["enrichment"]}` queries work.
- A long tail of correctness fixes found running this live in ed-dev — relation wiring direction (Content's own update vs Enrichment's create), null relationField filtering, stale-snapshot removal, schema field types (`language` string not array), transcript.json shape matching between human-edited and AI-generated captions — see individual commit messages.

## Test plan

- [x] Deployed and exercised live in ed-dev end-to-end (content publish ->
      transcription -> multilingual translation -> search)
- [ ] Verify in other environments before merge


content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+            ResponseHandler.OK
+              .put(ContentConstants.IDENTIFIER, contentIdentifier)
+              .put("transcriptId", transcriptNode.getIdentifier)
+              .put("message", "Transcription request accepted.")
+          }
+      }
+    }
+  }
+
+  private def createFromUpload(request: Request, contentNode: Node, enrichmentNode: Node)(implicit oec: OntologyEngineContext, ec: ExecutionContext, ss: StorageService): Future[Response] = {
```
sntiwari1, line 401: **Dead code path:** `isUpload` treats a `fileUrl`-only request as an upload (per the check at line 73: `containsKey("file") || containsKey("fileUrl")`), but `createFromUpload` only ever reads the `file` key. A client sending `fileUrl` without a multipart `file` will always hit `case _ => throw ClientException("ERR_MISSING_FILE", ...)` — this branch can never succeed as written.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+        // so contentId alone is already unique — the "_enrichment" suffix just makes
+        // it unambiguous in logs/dumps which kind of derived node this is, rather
+        // than duplicating the opaque generated identifier for no reason.
+        metadata.put("code", s"${contentNode.getIdentifier}_enrichment")
+        metadata.put("channel", channel)
+        metadata.put("contentId", contentNode.getIdentifier)
+        metadata.put("aiFeatures", new util.ArrayList[String]())
+        val createReq = buildTypedRequest(ENRICHMENT_OBJECT_TYPE, ENRICHMENT_SCHEMA_NAME, channel, metadata)
+        DataNode.create(createReq).flatMap { enrichmentNode =>
+          // Wires the edge from Content -> Enrichment. Must be added via a
```
sntiwari1, line 537: `DataNode.update(updateReq)` here is never chained/awaited before returning the success response — if the underlying update fails (validation/DB error), the client still gets `"Transcript updated."` with no error, since nothing observes this Future. Same pattern recurs around lines 383, 426/432, and 687-693 — worth chaining `.map`/`.flatMap` on the update result (or awaiting it) before responding OK.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+  // and omitted "id" entirely, which crashed that parser with KeyError('id').
+  private def buildTranscriptJson(segments: util.List[util.Map[String, AnyRef]]): String = {
+    val sb = new StringBuilder
+    sb.append("""{"segments":[""")
+    val segList = segments.asScala
+    segList.zipWithIndex.foreach { case (seg, idx) =>
+      val id = seg.getOrDefault("id", idx.asInstanceOf[AnyRef]).toString.toDouble.toInt
+      val start = seg.getOrDefault("start", "0").toString.toDouble
+      val end = seg.getOrDefault("end", "0").toString.toDouble
+      val text = escapeJson(seg.getOrDefault("text", "").asInstanceOf[String])
```
sntiwari1: `seg.getOrDefault("text", "")` only substitutes when the key is *absent* — an explicit `"text": null` segment still returns `null` here, and `escapeJson(null)` NPEs on `s.replace(...)`. This surfaces as an uncaught 500 instead of a validated 400. Consider `Option(seg.get("text")).getOrElse("")` or an explicit null check.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+      // definition). Enrichment's own config.json declares "transcripts" as
+      // direction "out", which is where this edge actually needs to
+      // originate from to pass validation.
+      //
+      // DefinitionNode.setRelationship treats the submitted relation list as
+      // the *complete* desired set for this relation name — any existing
+      // out-relation not resubmitted gets diffed into deletedRelations and
+      // removed. So every sibling Transcript's identifier must be resent
+      // here alongside the new one, or adding a second/third language wipes
+      // every previously-linked Transcript from Enrichment.transcripts.
```
sntiwari1, line 639: `kfClient.send` runs synchronously inside `.map` after the DB update has already succeeded (transcript flipped to Live), with no try/catch around it. If Kafka is unreachable, the Future fails and the client gets a 5xx for `approve` even though the state change was already persisted — a retry could then hit `ERR_TRANSCRIPT_NOT_IN_REVIEW`, masking that the original call actually succeeded. Consider isolating the Kafka publish in its own recover block so a broker hiccup doesn't turn a successful write into an error response.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+  private def resolveTargetTranscript(enrichmentNode: Node, transcriptId: String)(implicit oec: OntologyEngineContext, ec: ExecutionContext): Future[Node] = {
+    if (StringUtils.isBlank(transcriptId))
+      throw new ClientException("ERR_MISSING_TRANSCRIPT_ID", "transcriptId is required.")
+    readTranscriptChildren(enrichmentNode).map { transcripts =>
+      transcripts.find(t => StringUtils.equalsIgnoreCase(t.getIdentifier, transcriptId))
+        .getOrElse(throw new ClientException("ERR_TRANSCRIPT_NOT_FOUND", "No matching Transcript node found under this content's Enrichment."))
+    }
+  }
+
+  private def createTranscriptChildNode(contentIdentifier: String, enrichmentIdentifier: String, channel: String, languageCode: String,
```
sntiwari1, line 606: `createTranscriptChildNode`/`findOrCreateEnrichment` resubmit the full relation list from a snapshot read earlier in the same request, with no locking. Two concurrent `object/create` calls (e.g. different `languageCode`) for the same content can each read the same `existingChildIds` before either writes; per this file's own comment, unresubmitted out-relations get diffed into `deletedRelations` and removed, so the second write can silently drop the first request's newly-linked Transcript from `Enrichment.transcripts`.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
```diff
@@ -0,0 +1,939 @@
...
+          val uploadResult = ss.uploadFile(folderPath, vttFile, Option(false))
+          val captionsUrl = uploadResult(1)
+
+          val updateMetadata = new util.HashMap[String, AnyRef]()
+          updateMetadata.put("captionsUrl", captionsUrl)
+          updateMetadata.put("generatedBy", "human-uploaded")
+          updateMetadata.put("status", "Review")
+          val updateReq = buildTypedRequest(TRANSCRIPT_OBJECT_TYPE, TRANSCRIPT_SCHEMA_NAME, channel, updateMetadata)
+          updateReq.getContext.put("identifier", transcriptNode.getIdentifier)
+          DataNode.update(updateReq)
```
sntiwari1: The multipart VTT file copied to disk by the controller is uploaded to cloud storage here but never deleted locally — unlike `uploadTranscriptFiles`'s `finally { ...delete() }` pattern elsewhere in this same file. Repeated human VTT uploads via `object/create` will leave stray temp files accumulating on the content-service host.

content-api/content-controllers/src/main/scala/content/controllers/v4/ContentController.scala
```diff
@@ -170,6 +173,95 @@ class ContentController @Inject()(@Named(ActorNames.CONTENT_ACTOR) contentActor:
...
+
+    private def requestObjectFormData(identifier: String)(implicit request: Request[AnyContent]): java.util.Map[String, Object] = {
+        val reqMap = new java.util.HashMap[String, Object]()
+        request.body.asMultipartFormData.foreach { multipartData =>
+            multipartData.asFormUrlEncoded.foreach { case (key, values) =>
+                if (values.nonEmpty) reqMap.put(key, values.head)
+            }
+            if (multipartData.files.nonEmpty) {
+                val filePart = multipartData.files.head
+                val file = new java.io.File("/tmp" + java.io.File.separator + identifier + "_" + System.currentTimeMillis + "_" + filePart.filename)
```
sntiwari1: `requestObjectFormData` hardcodes `"/tmp"` for the upload temp path instead of reading it via `Platform`, even though `TranscriptManager` reads the conceptually-equivalent setting via `Platform.getString("content.upload.temp_location", "/tmp/content")`. Per `configuration-discipline.md`, paths like this should go through `Platform` with a sensible default so environments where `/tmp` isn't writable/desired can override it.

content-api/content-service/test/controllers/v4/ContentSpec.scala
```diff
@@ -35,7 +35,7 @@ class ContentSpec extends BaseSpec {
```
sntiwari1, line 35: None of the ~940 new lines in `TranscriptManager` or the 5 new `ContentActor` operations (create/update/approve/reject/readEnrichment) have corresponding test coverage — existing specs here were only touched to fix the `read()` signature. Per `testing.md`, normal/edge/failure paths should be covered before merge, especially given the status-transition and relation-wiring logic this PR introduces.

sntiwari1 (commented):
## Automated review summary

Reviewed via `/code-review` (effort: high). This PR adds a generic Enrichment/Transcript object model (schemas, ~940-line `TranscriptManager`, 5 new `ContentActor` operations, Kafka wiring) to support the ai-pipeline transcription/translation flow.

**Overall:** the feature design (generic Enrichment CRUD, per-object-type handler registry, opt-in `enrich` join on read) is solid, but the new `TranscriptManager` has several correctness gaps around discarded Futures, unguarded nulls, a dead code path, and a read-then-write race on relation updates. There's also no test coverage for any of the new logic. Details inline below.

### Findings
1. **Dead code / always-fails path** — `isUpload` accepts `fileUrl`-only requests, but `createFromUpload` only reads the `file` key, so that branch always throws `ERR_MISSING_FILE`.
2. **Discarded Futures** — `DataNode.update(...)` results aren't chained/awaited in several spots (`updateBySegmentEdit`, `createFromGeneration`, `createFromUpload`), so DB write failures are silently swallowed and callers get a success response regardless.
3. **Unchecked null → NPE** — `buildTranscriptJson`'s `escapeJson(seg.text)` NPEs on an explicit `"text": null` segment instead of failing with a `ClientException`.
4. **Kafka failure after successful DB write** — `pushEnrichedMetadataApprovedEvent`/`pushTranscriptionRequestEvent` run synchronously post-update with no try/catch; a Kafka outage turns an already-persisted approve into a client-visible 5xx.
5. **Read-then-write race on relations** — `createTranscriptChildNode`/`findOrCreateEnrichment` resubmit a snapshot of existing relations read earlier in the request; concurrent creates for the same content can silently drop each other's newly-linked Transcript.
6. **Local temp file never deleted** — the uploaded VTT file path in `createFromUpload` is never cleaned up, unlike the sibling `uploadTranscriptFiles` helper's `finally { ...delete() }`.
7. **No test coverage** — none of the ~940 new lines in `TranscriptManager` or the 5 new `ContentActor` operations have corresponding ScalaTest specs; existing specs were only touched to fix a `read()` signature change.
8. **Hardcoded temp path** — `requestObjectFormData` hardcodes `"/tmp"` instead of reading it via `Platform`, inconsistent with `TranscriptManager`'s own `content.upload.temp_location` config key.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Thanks! Fixed - now handles the fileUrl case too (downloads via SafeUrlValidator + FileUtils.copyURLToFile), and cleans up that downloaded temp file in a finally block.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Thanks! Fixed - this and the other unawaited DataNode.update calls (createFromGeneration's aiFeatures update, createFromUpload's two updates, syncEnrichmentTranscriptsFromNode's snapshot write, syncAndMaybeBuildEcar's ecarUrl write) now flatMap/await before proceeding.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Thanks! Fixed - Option(seg.get("text")).getOrElse("") in both buildTranscriptJson and buildVttContent now.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Thanks! Fixed - both push*Event methods now catch and log instead of propagating, so a broker hiccup can't turn an already-successful write into a 5xx.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Fixed the reads-from-a-stale-snapshot part: createTranscriptChildNode now re-reads the Enrichment node fresh immediately before writing the relation, instead of trusting existingChildIds from an earlier read in the same request. That narrows the race window significantly but doesn't fully eliminate it - two concurrent creates could still both re-read before either writes. Closing it completely needs real locking (per-Enrichment lock or optimistic-concurrency retry), which felt like a bigger change than this PR should carry - happy to open a follow-up if you'd rather track that separately.

content-api/content-actors/src/main/scala/org/sunbird/content/transcript/mgr/TranscriptManager.scala
chethann007: Thanks! Fixed - deletes the downloaded temp file in a finally block now (only the fileUrl-download branch's own temp file, not a real multipart upload's, which is the caller's to clean up).

content-api/content-controllers/src/main/scala/content/controllers/v4/ContentController.scala
chethann007: Thanks! Fixed - now reads content.upload.temp_location via Platform, same as TranscriptManager, defaulting to /tmp/content.

content-api/content-service/test/controllers/v4/ContentSpec.scala
chethann007: Added TranscriptManagerTest covering the pure helper functions (buildTranscriptJson, buildVttContent, formatVttTimestamp, escapeJson, toBool, isEcarReady) including regression cases for the null-text and isEcarReady bugs fixed alongside this - loosened those from private to private[mgr] to make them testable. The DataNode/graph/Kafka-dependent methods (create/update/approve/reject) would need a real mocking harness for Scala objects (PowerMock-style static mocking) that this codebase doesn't have yet - didn't want to bolt that on as a side effect of this fix. Flagging as a real gap rather than closing it silently; happy to scope that as follow-up work if useful.

pallakartheekreddy (approved):

PR #1311, opened 2026-08-20
feat: add artifactHash and prevArtifactHash fields to content schema

Deeksha1502:
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

### Type of change

Please choose appropriate options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes in the below checkboxes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Ran Test A
- [ ] Ran Test B

**Test Configuration**:
* Software versions: Java 11, scala-2.12, play-2.7.2
* Hardware versions: 2 CPU/ 4GB RAM

### Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

* **New Features**
  * Content records now support optional artifact integrity hashes.
  * Artifact hashes must use a valid 64-character lowercase hexadecimal format.

* **Improvements**
  * Content update operations can now copy artifact URLs, sizes, and artifact hash values.

### DIFF
--- a/schemas/content/1.0/schema.json
+++ b/schemas/content/1.0/schema.json
@@ -298,6 +298,14 @@
       "type": "string",
       "format": "url"
     },
+    "artifactHash": {
+      "type": "string",
+      "pattern": "^[a-f0-9]{64}$"
+    },
+    "prevArtifactHash": {
+      "type": "string",
+      "pattern": "^[a-f0-9]{64}$"
+    },
     "previewUrl": {
       "type": "string",
       "format": "url"
--- a/content-api/content-service/conf/application.conf
+++ b/content-api/content-service/conf/application.conf
@@ -534,7 +534,8 @@ content.copy.props_to_remove=["downloadUrl", "artifactUrl", "variants",
                                "LastPublishedBy", "rejectReasons", "rejectComment", "gradeLevel", "subject",
                                "medium", "board", "topic", "purpose", "subtopic", "contentCredits",
                                "owner", "collaborators", "creators", "contributors", "badgeAssertions", "dialcodes",
-                               "concepts", "keywords", "reservedDialcodes", "dialcodeRequired", "leafNodes", "sYS_INTERNAL_LAST_UPDATED_ON", "prevStatus", "lastPublishedBy", "streamingUrl"]
+                               "concepts", "keywords", "reservedDialcodes", "dialcodeRequired", "leafNodes", "sYS_INTERNAL_LAST_UPDATED_ON", "prevStatus", "lastPublishedBy", "streamingUrl",
+                               "artifactHash", "prevArtifactHash"]
 
 content.copy.origin_data=["name", "author", "license", "organisation"]
 
--- a/knowlg-service/conf/application.conf
+++ b/knowlg-service/conf/application.conf
@@ -604,7 +604,8 @@ content.copy.props_to_remove=["downloadUrl", "artifactUrl", "variants",
                                "LastPublishedBy", "rejectReasons", "rejectComment", "gradeLevel", "subject",
                                "medium", "board", "topic", "purpose", "subtopic", "contentCredits",
                                "owner", "collaborators", "creators", "contributors", "badgeAssertions", "dialcodes",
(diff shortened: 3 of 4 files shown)


schemas/content/1.0/schema.json
```diff
@@ -298,6 +298,14 @@
       "type": "string",
       "format": "url"
     },
+    "artifactHash": {
+      "type": "string",
+      "pattern": "^[a-f0-9]{64}$"
+    },
+    "prevArtifactHash": {
```
sntiwari1, line 305: **Integrity gap: `artifactHash`/`prevArtifactHash` will leak across content copy**

Verified in `CopyManager.getCopyRequest` (content-api/content-actors/src/main/scala/org/sunbird/content/util/CopyManager.scala:151-158): the copy path serializes *all* metadata off the source node and only strips the keys listed in `content.copy.props_to_remove` (content-api/content-service/conf/application.conf:529). That list already removes `artifactUrl` and `size` — but not `artifactHash`/`prevArtifactHash`.

Since a fresh `artifactUrl` is generated during copy (`CopyManager.artifactUpload`, ~line 292-309) but the old `artifactHash` is *not* stripped, a copied content node will end up with a brand-new artifact URL paired with the **previous artifact's hash**. For a field whose entire purpose is integrity verification, that's a real correctness/security bug, not just a schema nit.

Note this is separate from `config.json`'s `restrictProps.copy` — that list only blocks client-supplied request properties, it has no bearing on node-to-node metadata propagation during copy.

**Suggested fix:** add `"artifactHash", "prevArtifactHash"` to `content.copy.props_to_remove` in `content-api/content-service/conf/application.conf` (and let the artifact hash be recomputed post-upload, same as `artifactUrl`/`size`).

sntiwari1:
## Review summary

**Scope:** adds `artifactHash`/`prevArtifactHash` (64-char lowercase hex, optional) to the Content schema, and blocks them from being set directly on `updateContent` requests via `restrictProps.update`.

**Verified correct:**
- The `restrictProps.update` block-list is real and enforced: `RequestUtil.restrictProperties` is called from `ContentActor.update` (content-api/content-actors/.../ContentActor.scala:171-179), so a client can no longer spoof `artifactHash`/`prevArtifactHash` on a direct content update — important, since a client-controlled hash would defeat the point of an integrity field.
- Second commit correctly walked back the first commit's mistake of also adding `artifactUrl`/`size` to that same block-list (those two are fine to remain client-updatable, as before).
- Schema addition is backward compatible: not in the top-level `required` list, root object has no `additionalProperties: false`, so existing requests are unaffected.

**Real gap found (left inline on `schema.json`):** the copy path (`CopyManager.getCopyRequest`) propagates all source-node metadata except keys listed in `content.copy.props_to_remove` (content-api/content-service/conf/application.conf:529). That list strips `artifactUrl`/`size` but not the two new hash fields, so a copied content item gets a freshly uploaded artifact paired with the **previous** artifact's hash. Given these fields exist specifically for integrity verification, this should be fixed before merge — add `artifactHash`/`prevArtifactHash` to `props_to_remove`.

**Note (no action needed here):** grep across content-api/ontology-engine finds no code yet that computes or verifies `artifactHash` — these are pure schema additions in this PR. That's presumably landing in a follow-up commit on this branch (`feat/artifactHashing`); flagging only so the copy-path fix isn't forgotten once that hashing/verification logic lands.

pallakartheekreddy (approved):

PR #1313, opened 2026-09-15
fix(graph): let schema-typed String fields opt out of the JSON-array parse on read

likhithThammegowda:
### Problem

Any String property whose value starts with `[` and ends with `]` is parsed into an `ArrayList` when a node is read, regardless of the field's declared type in the content schema. This was introduced in df5eaac6 to restore genuinely list-typed properties after the Neo4j → JanusGraph migration, but it keys off the *shape* of the value rather than the schema.

A field the schema declares as `"type": "string"` is therefore returned as an array, and then fails its own validation on the next update:

```
PATCH /content/v3/update/<id>   {"request":{"content":{"zzProbe":"[{\"id\":\"x1\"}]"}}}   -> 200
GET   /content/v3/read/<id>     -> zzProbe comes back as an array, not the string that was written
```

Because validation runs on the **merged** node metadata, the node then rejects every subsequent update even when the payload does not carry that field at all — the content becomes permanently un-updatable. In our deployment this surfaced on `reviewer` (declared `string` in the content schema):

```
"Metadata reviewer should be a/an String value"
```

The same values also reach the search index, where these fields are mapped as `text`, and fail indexing with a `mapper_parsing_exception` — which cancels the `transaction-event-processor` Flink job and halts indexing for all content, not just the offending document.

### Fix

Two call sites both perform this conversion and both are needed — fixing only the first is not sufficient, because the node's metadata map then holds the String correctly and the response serializer converts it back to an array on the way out:

1. `JanusGraphNodeUtil.getNodeWithoutRelations` (graph-dac-api) — the graph read
2. `NodeUtil.convertJsonProperties` (graph-engine) — the read API response

Fields listed in the new `graph.string_only_fields` property are returned exactly as stored. The default is empty, so **behaviour is unchanged unless the property is configured**.

`DefinitionNode.convertJsonProperties` is deliberately untouched: it only runs on fields already filtered by the schema's `jsonProps`, which are genuinely JSON-typed.

### Configuration

```
graph.string_only_fields=["reviewer","creatorContacts","creatorDetails","publisherDetails","competencies_v1"]
```

### Verification

Deployed and verified on our staging environment:

- `content/v3/read` returns `reviewer` as a **string**
- a field *not* in the list still parses as an array, confirming the opt-out is selective rather than a blanket disable
- sending a course for review, and the reviewer → publisher transition, both work again

### Note

This is a targeted fix for the symptom. The more correct behaviour would be to drive the conversion from the schema's declared type rather than a configured list — happy to rework it that way if you would prefer.


likhithThammegowda:
Superseded by #1314, which carries the same four commits from the renamed branch `spark-v1.0.2.1-work`.

